PR #1, opened 2015-10-20
Add pure component transform

thejameskyle:
This is all I have time for now so I'm throwing it up into a PR.

~~Trying to figure out why Jest won't run.~~

Re: https://twitter.com/cpojer/status/656602748050247680

### DIFF
--- a/test/pure-component-test.output.js
+++ b/test/pure-component-test.output.js
@@ -1,0 +1,22 @@
+'use strict';
+
+var React = require('React');
+
+function render() {
+  return <div/>;
+}
+
+function Pure(props) {
+  return <div className={props.foo} />;
+}
+
+class Impure extends React.Component {
+  componentWillMount() {
+    // such impure
+  }
+  render() {
+    return <div className={this.props.foo} />;
+  }
+}
+
+var A = props => <div className={props.foo} />;
--- a/transforms/pure-component.js
+++ b/transforms/pure-component.js
@@ -1,0 +1,100 @@
+module.exports = function(file, api, options) {
+  const j = api.jscodeshift;
+  const ReactUtils = require('./utils/ReactUtils')(j);
+
+  const useArrows = options.useArrows || false;
+  const silenceWarnings = options.silenceWarnings || false;
+  const printOptions = options.printOptions || {
+    quote: 'single',
+    trailingComma: true
+  };
+
+  const getClassName = path =>
+    path.node.id.name;
+
+  const isRenderMethod = node => (
+    node.type == 'MethodDefinition' &&
+    node.key.type == 'Identifier' &&
+    node.key.name == 'render'
+  );
+
+  const onlyHasRenderMethod = path =>
+    j(path)
+      .find(j.MethodDefinition)
+      .filter(p => !isRenderMethod(p.value))
+      .size() === 0;
+
+  const THIS_PROPS = {
+    object: {
+      type: 'ThisExpression'
+    },
+    property: {
+      name: 'props'
+    }
+  };
+
+  const replaceThisProps = path =>
+    j(path)
+      .find(j.MemberExpression, THIS_PROPS)
+      .replaceWith(j.identifier('props'));
+
+  const buildPureComponentFunction = (name, body) =>
+    j.functionDeclaration(
+      j.identifier(name),
+      [j.identifier('props')],
+      body
+    );
+
+  const buildPureComponentArrowFunction = (name, body) =>
+    j.variableDeclaration(
+      'const', [
+        j.variableDeclarator(
+          j.identifier(name),
+          j.arrowFunctionExpression(
+            [j.identifier('props')],
+            body
+          )
+        )
+      ]
+    );
+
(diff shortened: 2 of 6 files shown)


zpao:
FWIW: tests look like they are running for me, but there are a few pre-existing failures

zpao:
Oh actually I bet it isn't wouldn't be running - add your case in https://github.com/reactjs/react-codemod/blob/master/test/__tests__/transform-tests.js. That won't solve general issues with jest not running but will make sure your test does run when you work that out.

cpojer:
yeah I'll take a look soon-ish at why the tests are failing.

@thejameskyle awesome work! :)

transforms/pure-component.js
```diff
@@ -0,0 +1,60 @@
...
+    [j.identifier('props')],
+    body
+  );
+}
+
+module.exports = function(file, api) {
+  const j = api.jscodeshift;
+  const f = j(file.source);
+
+  return j(file.source)
```
cpojer: This should do a `didTransform` kinda thing and only return a non-null value if the `.size()` of this collection is `> 0`.

See https://github.com/reactjs/react-codemod/blob/master/transforms/pure-render-mixin.js#L172

transforms/pure-component.js
```diff
@@ -0,0 +1,60 @@
...
+  return j.functionDeclaration(
+    j.identifier(name),
+    [j.identifier('props')],
+    body
+  );
+}
+
+module.exports = function(file, api) {
+  const j = api.jscodeshift;
+  const f = j(file.source);
```
cpojer: please copy pasta this: https://github.com/reactjs/react-codemod/blob/master/transforms/pure-render-mixin.js#L19

transforms/pure-component.js
```diff
@@ -0,0 +1,60 @@
...
+
+  return j(file.source)
+    .find(j.ClassDeclaration, REACT_COMPONENT_CLASS_DEFINITION)
```
cpojer: please filter that the class only has one method called `render` so it skips over more complex cases.

transforms/pure-component.js
```diff
@@ -0,0 +1,60 @@
...
+  object: {
+    type: 'ThisExpression'
+  },
+  property: {
+    name: 'props'
+  }
+};
+
+function replaceThisProps(node) {
+  return j(node)
```
cpojer: stupid question but did you test this? :P `j` is only part of the exported function below and not a global.

test/pure-component-test.output.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+var React = require('React');
+
+function render() {
+  return <div/>;
+}
+
+function A(props) {
```
cpojer: I still like the const + arrow function mode. If you wanna go crazy, how about making it an option for the hipster people? :)

cpojer:
This is a good start! The tests should be fixed on master now.

* Maybe take a look at the pure-render-mixin and class transforms in this file for the general structure. I think this codemod should be defensive and only inline actually pure components. It could print a `console.warn` that says that it skipped a class (`class.js` has a nice copy-paste-able warning thing).
* Move everything into `module.exports` to make it actually work
* Maybe add this: https://github.com/reactjs/react-codemod/blob/master/transforms/pure-render-mixin.js#L168 for safety!

transforms/pure-component.js
thejameskyle: I started refactoring and then tried to write tests and the tests didn't work and I didn't have time to look into it earlier.

fkling:
> It could print a console.warn that says that it skipped a class (class.js has a nice copy-paste-able warning thing).

There is also the `stats` function. It's basically a keyed counter. If you do a `dry` run, it prints how often `stats` was called with that argument. E.g. if a class with more than a render function is encountered, you could call `api.stats('complex class')` and it would print something like `complex class: 4` in the dry run.

Useful for collecting statistics about your codebase / codemod.

thejameskyle:
Threw together a Babel plugin as well. Thought it'd be a useful optimization to run in a production build https://github.com/thejameskyle/babel-plugin-react-pure-components

transforms/pure-component.js
```diff
@@ -0,0 +1,100 @@
...
+
+  pureClasses.replaceWith(p => {
+    const name = p.node.id.name;
+    const renderMethod = p.value.body.body.filter(isRenderMethod)[0];
+    const renderBody = renderMethod.value.body;
+
+    replaceThisProps(renderBody);
+
+    if (useArrows) {
+      return buildPureComponentArrowFunction(name, renderBody);
```
cpojer: this variant should probably take the `argument` of the `ReturnStatement` if the length of `renderBody.body` is one and it only contains a return statement :)

cpojer:
Yay! Would you mind flattening your commits into a single one before I merge? :)

cpojer:
Dammit, I missed this because github doesn't notify me about force pushes to PRs.

thejameskyle:
Sorry, I knew that, should've pinged you.

PR #4, opened 2015-10-21
Fixes #3

### LINKED ISSUE #3: Fix tests
Right now they don't pass. Looks like trailing commas are getting added by the printer.

zpao:


cpojer:
class-test2 seems broken to me, see https://github.com/reactjs/react-codemod/blob/master/test/__tests__/transform-tests.js#L57 which should not add the super class. Seems like something is wrong in class.js that I probably broke at some point. You mind removing those changes? I can look into them soon-ish.

zpao:
While we don't completely require that you extend React.Component yet, we're going to. So I think that's right.

cpojer:
Right, so in that case we should remove the option to not extend `React.Component` from class.js and also remove this test?

zpao:
Ah gotcha, didn't realize it was an option, just thought we changed code without updating the test. I'll remove those changes.

PR #9, opened 2015-10-29
Adding a codemod for transforming createElement function calls to JSX

TheSavior:
Transforms
```
var a = React.createElement(
  Foo,
  null,
  React.createElement("div", { foo: "bar" }),
  React.createElement(
    "span",
    null,
    "blah"
  )
);
```

to
```
var a = <Foo><div foo="bar"></div><span>blah</span></Foo>;
```

This codemod is useful for people who start out without using JSX and want to convert their codebase to JSX later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,10 @@ APIs.
 
 ### Included Scripts
 
+`create-element-to-jsx` converts calls to `React.createElement` into JSX elements.
+
+  * `jscodeshift -t react-codemod/transforms/create-element-to-jsx.js <file>`
+
 `findDOMNode` updates `this.getDOMNode()` or `this.refs.foo.getDOMNode()`
 calls inside of `React.createClass` components to `React.findDOMNode(foo)`. Note
 that it will only look at code inside of `React.createClass` calls and only
--- a/test/create-element-to-jsx-props.js
+++ b/test/create-element-to-jsx-props.js
@@ -1,0 +1,6 @@
+var React = require('React');
+
+function foo() {
+  var a = React.createElement(Foo, { foo: 'bar', bar: this.state.baz });
+  var b = React.createElement('div', { foo: 'bar', bar: this.state.baz });
+}
--- a/transforms/create-element-to-jsx.js
+++ b/transforms/create-element-to-jsx.js
@@ -1,0 +1,86 @@
+module.exports = function(file, api, options) {
+  const j = api.jscodeshift;
+  const root = j(file.source);
+  const ReactUtils = require('./utils/ReactUtils')(j);
+
+  const convertObjectExpressionToJSXAttributes = (objectExpression) => {
+    if (!objectExpression.properties) {
+      return [];
+    }
+
+    const attributes = objectExpression.properties.map((property) => {
+      if (property.type === 'SpreadProperty') {
+        return j.jsxSpreadAttribute(property.argument);
+      }  else if (property.type === 'Property') {
+        const propertyValueType = property.value.type;
+
+        let value;
+        if (propertyValueType === 'Literal') {
+          value = j.literal(property.value.value);
+        } else if (propertyValueType === 'MemberExpression') {
+          value = j.jsxExpressionContainer(property.value);
+        }
+
+        return j.jsxAttribute(
+          j.jsxIdentifier(property.key.name),
+          value
+        );
+      }
+    });
+
+    return attributes;
+  };
+
+  const convertNodeToJSX = (node) => {
+    const args = node.value.arguments;
+
+    const elementType = args[0].type;
+    const elementName = elementType === 'Literal' ? args[0].value : args[0].name;
+    const props = args[1];
+
+    const attributes = convertObjectExpressionToJSXAttributes(props);
+
+    const children = node.value.arguments.slice(2).map((child, index) => {
+      if (child.type === 'Literal') {
+        return j.literal(child.value);
+      }
+
+      return convertNodeToJSX(node.get('arguments', index + 2));
+    });
+
+    const openingElement = j.jsxOpeningElement(j.jsxIdentifier(elementName), attributes);
+
+    if (children.length) {
+      return j.jsxElement(
+        openingElement,
+        j.jsxClosingElement(j.jsxIdentifier(elementName)),
+        children
+      );
+    } else {
+      openingElement.selfClosing = true;
+      return j.jsxElement(openingElement);
+    }
+  };
+
+  if (ReactUtils.hasReact(root)) {
+    const mutations = root
+      .find(j.CallExpression, {
+        callee: {
+          object: {
+            name: 'React',
+          },
+          property: {
+            name: 'createElement',
+          },
+        },
+      })
+      .replaceWith(convertNodeToJSX)
+      .size();
+
+    if (mutations) {
+      return root.toSource();
+    }
+  }
+
+  return null;
+};
--- a/test/__tests__/create-element-to-jsx-test.js
+++ b/test/__tests__/create-element-to-jsx-test.js
@@ -1,0 +1,29 @@
+/**
+ * Copyright 2013-2015, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+'use strict';
+
+describe('create-element-to-jsx', () => {
+
+  it('transforms correctly', () => {
+    test('create-element-to-jsx', 'create-element-to-jsx-single-element');
+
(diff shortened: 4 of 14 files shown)


README.md
```diff
@@ -15,6 +15,10 @@ APIs.
 
 ### Included Scripts
 
+`create-element-to-jsx` converts calls to `React.createElement` into JSX elements.
```
TheSavior: Following alphabetical order

test/create-element-to-jsx-props.js
```diff
@@ -0,0 +1,4 @@
+function foo() {
+  var a = React.createElement(Foo, { foo: "bar", bar: this.state.baz });
+  var b = React.createElement("div", { foo: "bar", bar: this.state.baz });
```
cpojer: Can you add these two cases:

* `React.createElement(Foo, {foo: 'bar', ...someObject})` to `<Foo foo="bar" {...someObject}>` – where the order of properties should be preserved, of course.
* Convert children-less components to `<Foo />` instead of `<Foo></Foo>`

transforms/create-element-to-jsx.js
```diff
@@ -0,0 +1,69 @@
...
+   .forEach(update);
+
+  function update(node) {
+    j(node)
+    .replaceWith(
+      convertNodeToJSX(node)
+    );
+  }
+
+  return root.toSource();
```
cpojer: please move this to the bottom, after the function definitions.

transforms/create-element-to-jsx.js
```diff
@@ -0,0 +1,69 @@
+module.exports = function(file, api) {
+  var j = api.jscodeshift;
+  var root = j(file.source);
+
+  root.find(j.CallExpression, {callee: {object: {name: 'React'}, property: {name: 'createElement'}}})
```
cpojer: can you make this fit into an 80 column limit + add trailing commas?

(I think I have to add eslint to this repo :) )

transforms/create-element-to-jsx.js
```diff
@@ -0,0 +1,69 @@
...
+   .forEach(update);
+
+  function update(node) {
+    j(node)
+    .replaceWith(
```
cpojer: indentation! :)

```js
  j(node)
    .replaceWith(
      …
    );
```

or

```js
  j(node).replaceWith(…);
```

here and everywhere else in the file please :)

cpojer:
This is awesome! I left a couple of inline comments and I'll set up eslint for this repo.

Two more minor things:

* Can you copy-paste the way printOptions works in the pure-render-transform: https://github.com/reactjs/react-codemod/blob/master/transforms/pure-render-mixin.js#L19 – I need to turn this pattern into some utility function or something, this is really annoying to duplicate :(
* If a transform doesn't touch a file it should `return null` instead of `root.toSource()` to avoid unwanted mutations. The other transforms in this repo all count how many things they transform and only return something if they find callers they want to mutate.

test/create-element-to-jsx-props.js
TheSavior: Is it correct that 
```
var a = React.createElement(Foo, null);
var b = React.createElement("div", null);
```

should convert to

```
var a = <Foo />;
var b = <div></div>;
```
then?

cpojer: I think they should both convert to a single tag that closes, like `<div />`.

transforms/create-element-to-jsx.js
TheSavior: ESlint isn't currently enforcing trailing commas or the 80 column limit.

TheSavior: Ah, it does enforce trailing commas, but only if things are on multiple lines (which they will be)

TheSavior:
[] Create test for the spread operator
 [] Return null if file wasn't modified

cpojer:
oh yeah, can you also flatten your commits into a single one?

transforms/create-element-to-jsx.js
```diff
@@ -0,0 +1,95 @@
...
+      );
+    } else {
+      openingElement.selfClosing = true;
+      return j.jsxElement(openingElement);
+    }
+  };
+
+  const update = (node) => {
+    j(node)
+      .replaceWith(
```
cpojer: you can just inline this `replaceWith` below:

`root.find(...).replaceWith(...).size()`

TheSavior: As in combining this and the stuff on line 76?

transforms/create-element-to-jsx.js
```diff
@@ -0,0 +1,95 @@
...
+          value
+        );
+      }
+    });
+
+    return attributes;
+  };
+
+  const convertNodeToJSX = (node) => {
+    if (!node.value || node.value.type !== 'CallExpression') {
```
cpojer: I don't think you need this, right? Because you are returning undefined it would remove the node and the `find` call makes sure it is safe to call `convertNodeToJSX`. I'd prefer to put the guard around line #52

transforms/create-element-to-jsx.js
cpojer: yeah, you don't need the call to `j`, `replaceWith` is part of every collection so you can just call it directly instead of `forEach`.

cpojer:
I like no-parens around single-argument-identifiers in arrows but I'm definitely not going to fight you over that :P @zpao will love you, actually, so this is just about who you wanna score points with.

transforms/create-element-to-jsx.js
TheSavior: Neat.

transforms/create-element-to-jsx.js
TheSavior: Yeah, this check is left over from a weird recursive call I had before I learned about `.get()` for getting paths.

TheSavior:
If single-argument-identifiers was enforced by eslint I'd be happy to follow it.  :)

transforms/create-element-to-jsx.js
```diff
@@ -0,0 +1,95 @@
...
+  const j = api.jscodeshift;
+  const root = j(file.source);
+
+  const convertObjectExpressionToJSXAttributes = (objectExpression) => {
+    if (!objectExpression.properties) {
+      return [];
+    }
+
+    const attributes = objectExpression.properties.map((property) => {
+
```
cpojer: rm this empty line please :)

transforms/create-element-to-jsx.js
```diff
@@ -0,0 +1,95 @@
...
+      if (property.type === 'SpreadProperty') {
+        return j.jsxSpreadAttribute(property.argument);
+      }  else if (property.type === 'Property') {
+        const propertyValueType = property.value.type;
```
cpojer: you can also use destructuring here:

```js
const {type} = property.value;
```

transforms/create-element-to-jsx.js
```diff
@@ -0,0 +1,95 @@
...
+
+        let value;
```
cpojer: and if you make the variable name shorter you can probably just use a ternary here :P

transforms/create-element-to-jsx.js
```diff
@@ -0,0 +1,95 @@
...
+
+    return attributes;
+  };
+
+  const convertNodeToJSX = (node) => {
+    if (!node.value || node.value.type !== 'CallExpression') {
+      return undefined;
+    }
+
+    const args = node.value.arguments;
```
cpojer: `const {arguments: args} = node.value;` if you wanna go nuts.

TheSavior: Hah, I'll resist unless you prefer it.

cpojer:
Alright this looks really solid. I left a couple of inline comments but take your pick on what you actually wanna update – these are mostly just suggestions.

The only thing I'd like to see is to copy this: https://github.com/reactjs/react-codemod/blob/master/transforms/class.js#L500 to make sure it only runs on code that actually has React in scope and doesn't have a `React = require('underscore')` or something in it :)

TheSavior:
Can you give me a test case that would fail without that check?

cpojer:
It should be something like doing `var React = require('foo')`, like you map React to something that is not actually React.

cpojer:
I also promise to consolidate all that setup code some time, it sucks to copy-paste it.

TheSavior:
Currently failing for the case of 

```
var React = require('foo');

React.createElement(Foo, 'la');
```

```
Expected: '' toEqual: 'var React = require('foo');
    
    React.createElement(Foo, 'la');'

```

Should my test expect that to return the original contents untransformed, or an empty string as it is now? 

Also, I don't see any tests for any of the other transforms that test the behavior of `options['explicit-require']` so I don't have examples to check for this.

cpojer:
yeah it is ok not to test it. If the transform returns null it means that it won't apply any transformations, so you should test for an empty string (and maybe add the coercion from null to an empty string in `jest/env.js`.

cpojer:
yay!

PR #10, opened 2015-11-08
create-element-to-jsx-children: Convert .map calls into a JSXExpressi…

TheSavior:
…onContainer

This didn't properly handle converting:
```
var React = require('React');

React.createElement(
  'div',
  null,
  foo.map(function() {})
);
```

to:

```
var React = require('React');

<div>{foo.map(function() {})}</div>;
```

Now it does! :)


cpojer:
nice!

PR #12, opened 2015-11-08
create-element-to-jsx-children: Convert mixed strings to JSXText instead of Literal

TheSavior:
Based on #10 

This

```
React.createElement(
  'div',
  null,
  a,
  ' ',
  a
);
```

was getting transformed to

```
<div>{a}" "{a}</div>;
```

Now there are no unnecessary quotes.

However, I'm having a problem.
This
```
React.createElement(
  'div',
  null,
  ' lala '
);
```
is being converted to
```
<div>lala</div>;
```

Losing the padding spaces. It seems like `j.jsxText('  foo  ');` gets output as `foo`. `literal` has the same issue. So how can I make that case work?

### DIFF
--- a/test/__tests__/create-element-to-jsx-test.js
+++ b/test/__tests__/create-element-to-jsx-test.js
@@ -27,6 +27,8 @@ describe('create-element-to-jsx', () => {
 
     test('create-element-to-jsx', 'create-element-to-jsx-children-map');
 
+    test('create-element-to-jsx', 'create-element-to-jsx-children-mixed-empty-string');
+
     test('create-element-to-jsx', 'create-element-to-jsx-spread');
 
     test('create-element-to-jsx', 'create-element-to-jsx-no-react');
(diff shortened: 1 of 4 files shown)


cpojer:
Oh this kinda sucks. It might require you to look back/look ahead?

TheSavior:
I'm not sure what you mean by looking back or looking ahead, but it seems like looking at surrounding nodes wouldn't solve the problem of `jsxText` / `literal` trimming their input.

TheSavior:
I just removed the test for the padded string. It was conflating the other fix of `<div>{a}" "{a}</div>;` getting output. Let's tackle that one separately.

cpojer:
Oh, sorry, I thought fixing the first thing would break the second and vice-versa. So you are saying that `j.jsxText(' foo ')` always prints as `foo` instead of ` foo `.

TheSavior:
Yes, that is what I'm saying. :)

cpojer:
You might have to take a look into this: https://github.com/benjamn/recast/blob/master/lib/printer.js#L1089 and how fromString works.

TheSavior:
Let's get this one merged and I'll take a look at that.

cpojer:
sure!

PR #15, opened 2015-11-15
Handling props as a spread attribute

### LINKED ISSUE #14: create-element-to-jsx doesn't handle props objects
The create-element-to-jsx transform attempts to make replacements like this:

~~~.diff
-    return React.createElement(Constructor, props);
+    return <Constructor />;
~~~

Expected:

~~~.diff
-    return React.createElement(Constructor, props);
+    return <Constructor {...props} />;
~~~

TheSavior:
Fixes #14


cpojer:
Nice. Thanks Eli!

PR #18, opened 2015-12-09
Allow for a callExpression as child that's not an element.

RSO:
The previous code assumed that all children that were of type
CallExpression were in fact calls on MemberExpressions. This fix makes
sure we also support other calls.


cpojer:
Can you rebase this? It conflicts with your other pull request that I just merged :)

RSO:
Sure thing!

PR #19, opened 2015-12-10
Support React.__spread/Object.assign in JSX transform

RSO:
While reviewing the output of the transform I noticed that the transform lost all attributes for elements that use the `React.__spread` helper. I decided to fix this and make the transform throw an error when it doesn't recognise a certain attribute. This way we can be more sure that things don't fall through.

### DIFF
--- a/test/__tests__/create-element-to-jsx-test.js
+++ b/test/__tests__/create-element-to-jsx-test.js
@@ -44,4 +44,16 @@ describe('create-element-to-jsx', () => {
     test('create-element-to-jsx', 'create-element-to-jsx-object-assign');
   });
 
+  it('raises when it does not recognize a property type', () => {
+    const jscodeshift = require('jscodeshift');
+    const transform = require('../../transforms/create-element-to-jsx');
+    const source = `
+      var React = require("react/addons");
+      React.createElement("foo", 1)
+    `;
+
+    expect(() => transform({source}, {jscodeshift}, {}))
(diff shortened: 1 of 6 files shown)


RSO:
I should probably rebase one of these PRs once the other is landed.

cpojer:
yeah, I'm sorry, you need to rebase again :) Thanks for the contribution, may I ask where you are using this? :)

RSO:
Not a problem! 

I've been actively investigating transforming our (@Hacker0x01) codebase from CoffeeScript (CJSX) to ES2015 by using transforms to make the ES5 output ES2015. Currently most of my work lives in https://github.com/Hacker0x01/espresso, but I have plans of removing all the CLI helper stuff and making it a repository filled with transforms (and tests), just like this repository. I want to do this because I plan on running all transforms one by one to make sure there aren't any unforeseen cases that may break (as the assumptions on which some transforms are based are pretty loose).

I'm demoing the outcome of my investigations to my team next week, so I'll know after that if/when we decide to go forward with this :-).

Thank you for your amazing JSConf talk, it was the spark that ignited this whole project!

cpojer:
oh man, this is really awesome. We talked about building a CS to JS converter, also using jscodeshift, mainly just out of hate ;) Keep up the good work!

fkling:
@RSO: This looks great! Also let us know if you have any suggestions for jscodeshift.

PR #20, opened 2015-12-23
Add info about requiring React

ustun:
If React is not required, the transformations ignore the target files. Added a note to clarify this issue.


cpojer:
For some reason I missed this PR. Thanks for the contribution but all codemods have a `--no-explicit-require` for this use-case so I don't think we need the documentation :)

ustun:
Oh, I had missed that. Great, closing.

PR #21, opened 2015-12-31
add sort-comp transform into compliance with eslint react/sort-comp

turadg:
Reorders React component methods to match the [ESLint](http://eslint.org/) [react/sort-comp rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md), specifically with the [tighter constraints of the Airbnb style guide](https://github.com/airbnb/javascript/blob/6c89f9587f96688bf0d4d536adb8eb4ed9bf0002/packages/eslint-config-airbnb/rules/react.js#L47-L57):
```
 'react/sort-comp': [2, {
   'order': [
     'lifecycle',
     '/^on.+$/',
     '/^(get|set)(?!(InitialState$|DefaultProps$|ChildContext$)).+$/',
     'everything-else',
     '/^render.+$/',
     'render'
   ]
 }],
```

NOTE: only works on React.createClass() syntax, not ES6 class.

### DIFF
--- a/transforms/sort-comp.js
+++ b/transforms/sort-comp.js
@@ -1,0 +1,141 @@
+/**
+ * Reorders React component methods to match the [ESLint](http://eslint.org/)
+ * [react/sort-comp rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md),
+ * specifically with the [tighter constraints of the Airbnb style guide](https://github.com/airbnb/javascript/blob/6c89f9587f96688bf0d4d536adb8eb4ed9bf0002/packages/eslint-config-airbnb/rules/react.js#L47-L57),
+ *
+ *  'react/sort-comp': [2, {
+ *    'order': [
+ *      'lifecycle',
+ *      '/^on.+$/',
+ *      '/^(get|set)(?!(InitialState$|DefaultProps$|ChildContext$)).+$/',
+ *      'everything-else',
+ *      '/^render.+$/',
+ *      'render'
+ *    ]
+ *  }],
+ *
+ * NOTE: only works on React.createClass() syntax, not ES6 class.
+ */
+module.exports = function(fileInfo, api, options) {
+  const j = api.jscodeshift;
+
+  const ReactUtils = require('./utils/ReactUtils')(j);
+
+  const printOptions =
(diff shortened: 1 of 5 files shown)


transforms/sort-comp.js
```diff
@@ -0,0 +1,147 @@
...
+      // compare by index
+      return indexA - indexB;
+    }
+  };
+
+  const sortComponentProperties = classPath => {
+    const spec = ReactUtils.getReactCreateClassSpec(classPath);
+
+    if (spec) {
+      const orderedProperties = spec.properties.sort(propertyComparator);
```
cpojer: I think if you remove the lines below and you just do `spec.properties.sort(propertyComparator)` it should be enough and mutate the nodes in-place.

cpojer:
Cool! Would you mind adding a test and making the change for the inline-comment? :)

transforms/sort-comp.js
turadg: `spec` is the contents of the `createClass`, so if I don't `replaceWith` a call to `ReactUtils.createCreateReactClassCallExpression` then I end up with,
```
-const App = React.createClass({
+const App = {
```

I could still sort in place, but I think it makes the code less clear because if your eyes jump to the line where `spec` is passed as an arg, then you have to read back to see that it has been mutated.

```
    if (spec) {
      spec.properties.sort(propertyComparator);

      ReactUtils.findReactCreateClassCallExpression(classPath).replaceWith(
        ReactUtils.createCreateReactClassCallExpression(
          spec.properties
        )
      );
    }
```

If that's a condition to accepting this I'll make the change, but I do think it's not an improvement.

cpojer: the idea for codemods should be to only make the minimal changes necessary, which is why I asked you to make this change. JavaScript's `sort` is in-place, meaning that `spec.properties.sort(propertyComparator);` should mutate the original array and re-printing the AST should yield the correct result.

turadg: Apologies, it didn't appear to work when I tried it. I see now that it works, and is better because it doesn't make additional changes. On the plus side, while debugging that I caught a different bug. :) (Also fixed)

cpojer:
awesome! Seems good to me. Thanks for the contribution and keep it up! Feel free to join us on #codemod on reactiflux: www.reactiflux.com

PR #23, opened 2016-01-11
Make sure less-than and great-than characters are properly encoded.

RSO:
Previously the transform would translate

```
React.createElement("div", null, "\x3C");
```

To

```
<div><</div>;
```

This PR fixes this by translating `<>` to `&lt;&gt;`.

### DIFF
--- a/test/create-element-to-jsx-gt-lt-entities.js
+++ b/test/create-element-to-jsx-gt-lt-entities.js
@@ -1,0 +1,3 @@
+var React = require('React');
+
+React.createElement('div', null, '\x3C\x3E');
--- a/transforms/create-element-to-jsx.js
+++ b/transforms/create-element-to-jsx.js
@@ -4,13 +4,8 @@ module.exports = function(file, api, options) {
   const ReactUtils = require('./utils/ReactUtils')(j);
   const encodeJSXTextValue = value =>
     value
-      .replace(new RegExp('(?:<|>)', 'g'), match => {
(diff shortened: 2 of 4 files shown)


test/create-element-to-jsx-gt-lt-entities.js
```diff
@@ -0,0 +1,3 @@
+var React = require('React');
+
+React.createElement('div', null, '\x3C\x3E');
```
RSO: `\x3C\x3E` is how our CJSX compiler ends up outputting to ES5. I'm not sure if I should add support for other ways of writing these characters as well.

RSO:
Not sure about the failure, can't reproduce locally.

transforms/create-element-to-jsx.js
```diff
@@ -2,6 +2,15 @@ module.exports = function(file, api, options) {
   const j = api.jscodeshift;
   const root = j(file.source);
   const ReactUtils = require('./utils/ReactUtils')(j);
+  const encodeJSXTextValue = value =>
+    value
+      .replace(new RegExp('(?:<|>)', 'g'), match => {
```
cpojer: Can you turn this into to individual regexes? Seems way simpler:

```
value.replace(/</g, '&lt;').replace(/>/g, '&gt;');
```

cpojer:
thanks! Would you mind rebasing now that I merged your other changes? :)

PR #24, opened 2016-01-11
Allow for MemberExpression in JSXIdentifier

RSO:
Tests should cover why I would like to have this.

### DIFF
--- a/transforms/create-element-to-jsx.js
+++ b/transforms/create-element-to-jsx.js
@@ -66,20 +66,23 @@ module.exports = function(file, api, options) {
     }
   };
 
-  const jsxIdentifierNameFor = node => {
+  const jsxIdentifierFor = node => {
     if (node.type === 'Literal') {
-      return node.value;
+      return j.jsxIdentifier(node.value);
     } else if (node.type === 'MemberExpression') {
-      return `${jsxIdentifierNameFor(node.object)}.${jsxIdentifierNameFor(node.property)}`;
+      return j.jsxMemberExpression(
+        jsxIdentifierFor(node.object),
+        jsxIdentifierFor(node.property)
+      );
     } else {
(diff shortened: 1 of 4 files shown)


transforms/create-element-to-jsx.js
```diff
@@ -60,11 +60,20 @@ module.exports = function(file, api, options) {
     }
   };
 
+  const jsxIdentifierNameFor = node => {
```
cpojer: Can we change this to just return the node itself instead of generating a string?

RSO: Doing that would make the recursion harder to accomplish. The recursion is needed to allow for nested MemberExpressions.

cpojer: why do you need the recursion? Can't you just return "node" if its a MemberExpression and make the code accept either an AST node or a string?

transforms/create-element-to-jsx.js
```diff
@@ -66,11 +66,23 @@ module.exports = function(file, api, options) {
     }
   };
 
+  const jsxIdentifierFor = node => {
```
RSO: @cpojer as you can see, the method now returns a node, but the recursion is still there, as we need to be able to convert `React.createElement(A.b.c.d)` as well. So `node.object` can be another `MemberExpression` that needs to be converted to a `JsxMemberExpression`, and `node.object` is an Identifier that should be converted to a `jsxIdentifier`. I could probably just change that for `j.jsxIdentifier(node.name)` but that would technically be duplicating the `else` block.

I hope I understood your request ;-) let me know what you think!

cpojer: sounds good! I honestly didn't know that jsxMemberExpression was something different from a MemberExpression, that's silly.

PR #25, opened 2016-01-11
Allow for more prop types to translate to spread

RSO:


transforms/create-element-to-jsx.js
```diff
@@ -14,9 +14,13 @@ module.exports = function(file, api, options) {
       expression.callee.object.name === 'Object' &&
       expression.callee.property.name === 'assign';
 
-    if (expression.type === 'Identifier') {
-      return [j.jsxSpreadAttribute(expression)];
-    } else if (isReactSpread || isObjectAssign) {
+    const validSpreadTypes = [
```
cpojer: this should probably be an object so that lookups are faster. It doesn't matter in this case so I'm gonna merge it :)

PR #26, opened 2016-01-20
@jnwng

lewisf:
This adjusts the class.js codemod to do es7 static class properties. Not sure how to test
this since it requires external libraries so i was unable to use the in browser analyzer.
Trying to see if facebook has an atom plugin or something.

Future improvements:
- Fat arrow function instead of .bind(this) in constructor
- Sorting of functions based on our linter configuration.


lewisf:
sorry, pull request on the wrong repo

PR #28, opened 2016-02-04
Add support for ES6 classes in sort-comp

### LINKED ISSUE #27: Add support for ES6 classes in sort-comp
In a comment in sort-comp, I noticed the following comment:

> NOTE: only works on React.createClass() syntax, not ES6 class.

It would be really lovely if this worked with ES6 classes. :smiley_cat:

lencioni:
In a comment in sort-comp, I noticed the following comment:

> NOTE: only works on React.createClass() syntax, not ES6 class.

I thought that it would be really lovely if this worked with ES6
classes, so I decided to take a crack at adding support.

Fixes #27


cpojer:
Nice work! Thank you so much for your contribution.

PR #31, opened 2016-02-26
Teach sort-comp to put static methods at the top

lencioni:
The Airbnb JavaScript style guide was recently updated to sort static
methods to the tops of classes [0]. Since this codemod is designed to
follow the constraints of the Airbnb style guide, I figured it would be
good to update this as well.

[0]: https://github.com/airbnb/javascript/pull/715

### DIFF
--- a/transforms/sort-comp.js
+++ b/transforms/sort-comp.js
@@ -1,10 +1,11 @@
 /**
  * Reorders React component methods to match the [ESLint](http://eslint.org/)
  * [react/sort-comp rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md),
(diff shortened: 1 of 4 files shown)


cpojer:
thank you!

transforms/sort-comp.js
```diff
@@ -119,24 +121,40 @@ const regExpRegExp = /\/(.*)\/([g|y|i|m]*)/;
...
   var j;
   var indexes = [];
-  for (i = 0, j = methodsOrder.length; i < j; i++) {
-    isRegExp = methodsOrder[i].match(regExpRegExp);
-    if (isRegExp) {
-      matching = new RegExp(isRegExp[1], isRegExp[2]).test(method);
-    } else {
-      matching = methodsOrder[i] === method;
+  // Check for static methods
+  if (indexes.length === 0) {
```
turadg: indexes.length has to equal zero at this point. it just initialized to empty array in the line above.

lencioni: Indeed. I decided to add the check here to make it so we don't have to change this again if we add something before it.

turadg: Ah. What do you think of #32?

PR #32, opened 2016-02-26
simplify getRefPropIndexes() logic

turadg:
I noticed some extra loop iteration in #31 so did a little refactoring.

[Review on Reviewable](https://reviewable.io/reviews/reactjs/react-codemod/32)


cpojer:
Oh, thanks for the cleanup! :)

PR #33, opened 2016-02-27
take method ordering in sort-comp as an option

turadg:
Also simplifies the logic. I originally based it off https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/sort-comp.js, but it doesn't need to be nearly as complex because it doesn't have to give useful error output.

[Review on Reviewable](https://reviewable.io/reviews/reactjs/react-codemod/33)


cpojer:
this is awesome! Thank you so much :)

Issue #34, opened 2016-02-27
Create JSX - file skipped with no explanation


chrismcv:
Hi, I'm playing with this for this first time.  My test file is being skipped, and nothing is being logged as to why.

```
jscodeshift -t react-codemod/transforms/create-element-to-jsx.js -p --dry --run-in-band -v 2 output.js 
Processing 1 files...
Running in dry mode, no files will be written!
 SKIP output.js
All done.
Results: 0 errors 0 unmodified 1 skipped 0 ok
Time elapsed: 0.999 seconds
```

Is there any way I can get more output from it?

cpojer:
Can you paste the file that is being skipped?

chrismcv:
https://gist.github.com/chrismcv/b0de8c90337e9bdce9a1

cpojer:
I guess you can comment out https://github.com/reactjs/react-codemod/blob/master/transforms/create-element-to-jsx.js#L121 Other codemods in this repo have a `--no-explicit-react` flag, see https://github.com/reactjs/react-codemod/blob/master/transforms/findDOMNode.js#L128 – feel free to send a pull request that adds this to the create-element-to-jsx codemod.

chrismcv:
Thanks - that was it. PR submitted

PR #35, opened 2016-02-27
[Create-Element-To-JSX] support for --no-explicit-react

chrismcv:
As discussed in #34


cpojer:
Nice! Thanks for the contribution :)

PR #37, opened 2016-03-24
Update the README to follow jscodeshift doc

oliviertassinari:
Signature according to https://github.com/facebook/jscodeshift#usage-cli is actually `<path>`.


cpojer:
Sure! Thank you.

PR #38, opened 2016-04-04
Add codemod to add px values to end of style properties using shorthand syntax

conorhastings:
Adapting from my repo here https://github.com/conorhastings/react-style-px-suffix-codemod 

@cpojer please let me know if there is anything else i can do!


README.md
```diff
@@ -101,6 +101,15 @@ guide](https://github.com/airbnb/javascript/blob/7684892951ef663e1c4e62ad57d662e
 jscodeshift -t react-codemod/transforms/sort-comp.js <path>
 ```
 
+#### `append-px-to-style-properties`
+
+Updates code to get rid of warning in React 15 when using shorthand style syntax. Future versions of React will not automatically append `px` to the end of values specified with shorthand. Optionally you can pass comma delimited values to `ignore` on command line for style properties you want to specifically not transform.
```
cpojer: `get rid of warnings`?

cpojer: and `on the command line`.

test/append-px-to-style-properties-basic-transform.output.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+class Cool extends React.Component {
+  render() {
+    const style = { height: '10px', width: '10px', flex: 1 };
+    const style2 = { height: '20px', width: '20px' };
+    return <div style={{fontSize: '10px'}}><div style={style}>Hello</div><div style={style2}>Hello</div></div>;
+  }
+}
```
cpojer: would you mind adding a newline at the end of every file? :)

transforms/append-px-to-style-properties.js
```diff
@@ -0,0 +1,134 @@
...
+  } else if (type === 'ConditionalExpression') {
+    transformConditional(style, root);
+  }
+} 
+
+module.exports = function(file, api, options) {
+  if (options.ignore) {
+    const toIgnore = options.ignore.split(',');
+    if (toIgnore.length) {
+      itemsToNotSuffixWithPx = itemsToNotSuffixWithPx.concat(toIgnore);
```
cpojer: I think this will be cleaner if you put all code into the exported function. Then you can create a local variable of items that shouldn't be suffixed instead of mutating one global variable.

transforms/append-px-to-style-properties.js
```diff
@@ -0,0 +1,134 @@
...
+  'strokeDashoffset',
+  'strokeOpacity',
+  'strokeWidth',
+  'tabSize',
+  'widows',
+  'zIndex',
+  'zoom',
+];
+
+const shouldSuffixIfIntergerValue = attr => itemsToNotSuffixWithPx.indexOf(attr) === -1;
```
cpojer: you'll be able to make your code a bit faster if you turn `itemsToNotSuffixWithPx` into an object (or a `Set`) – lookups will be faster than using `indexOf` on an array.

transforms/append-px-to-style-properties.js
```diff
@@ -0,0 +1,134 @@
...
+      transformStyleObject(p.argument, root);
+    } else {
+      const key = p.key.type === 'Identifier' ? p.key.name : p.key.value;
+      let val = p.value.value;
+      if (val === undefined) {
+        if (p.value.type === 'UnaryExpression') {
+          val = parseInt(`${p.value.operator}${p.value.argument.value}`, 10); 
+          if (shouldSuffixIfIntergerValue(key) && typeof val === 'number') {
+            p.value.operator = '';
+            p.value.argument.value = `${val}px`;
```
cpojer: I'm surprised that printing this works, but I guess a `UnaryExpression` without an operator just prints the value?

cpojer:
@zpao @balpert can you take a look at this? If it's good, we should announce it together with the v15 release blog post :)

spicyj:
AFAIK we're only requiring px on _string_ style values like `width: '150'`, but still auto-suffixing numbers like `width: 150`.

conorhastings:
@spicyj ahh didn't catch that, i can update it to just look for those if that's desirable.

spicyj:
Sure. I would guess that it's rare string values come from literals though; they're probably usually more indirect.

conorhastings:
@spicyj @cpojer updated transform to only transform strings without px value by default, with optional flag to also transform `int` values. Let me know what you think about that being a potential option

PR #39, opened 2016-04-19
Account for exports.

mareksuscak:
This PR adds the support for ES6 exports in the findDOMNode transform.


cpojer:
Thank you for your contribution! :)

PR #40, opened 2016-04-27
Move unit tests to new jscodeshift test util

Daniel15:
Moves all the unit tests in this project to the [new jscodeshift test util](https://github.com/facebook/jscodeshift/pull/104) I added recently. This avoids the need to use the global "test" function that was previously defined in `jest/env.js`

I've still kept the test function and just added a deprecation notice to it, as in theory other projects //may// be using it. The only [public repo I could find on Github that was using it](https://github.com/search?utf8=%E2%9C%93&q=node_modules%2Freact-codemod%2Fjest%2Fenv.js&type=Code&ref=searchresults) is @cpojer's js-codemod, which I've already submitted a diff for (https://github.com/cpojer/js-codemod/pull/36). I'm just worried that other projects are doing the exact same thing, especially if they used js-codemod as a 'template'. We can delete it eventually.


cpojer:
Thank you!

PR #50, opened 2016-06-02
Fix create-element-to-jsx failing when props argument is missing

valscion:
`React.createElement(Component)`, without the second `props` argument,
was failing with `TypeError: Cannot read property 'type' of undefined`
error. As `React.createElement()` allows emitting the `props` argument
altogether as far as I have understood, the codemod shouldn't fail when
props argument is indeed missing.


cpojer:
Thank you! :)

PR #52, opened 2016-06-02
51: handle failure to read eslint config options for file

reergymerej:
adds try...catch around reading of eslintrc in sort-comp


cpojer:
yeah that makes sense! Thanks.

PR #57, opened 2016-06-15
56: identify components with refs as impure

### LINKED ISSUE #56: pure-component does not check for refs
A component can only be converted into a stateless component if it does not use refs on its children. The current codemod does not check for it.
Still amazed by these codemods, but this could be a great improvement...

reergymerej:
This fixes #56, skips over components that render with a `ref`.


cpojer:
Thanks!

PR #58, opened 2016-06-21
Upgrade eslint and jest

keyanzhang:
cc @cpojer


cpojer:
Thanks! Can you add `"testEnvironment": "node"` to the Jest config? It should make Jest's startup instant because we won't have to wait for jsdom.

keyanzhang:
Sure!

Issue #59, opened 2016-06-26
create-element-to-jsx doesn't preserve comments


alangpierce:
It looks like the `create-element-to-jsx` transform silently throws away comments within `React.createElement` calls. Ideally, it should find some way to preserve the comments, or maybe give an error so I can fix those files manually, but if that's too hard or out of scope, I think the behavior should at least be documented.

For example, if you run it on this file:
```js
import React from 'react';

const render = () => {
  return React.createElement('div', {
    className: 'foo',  // This is an important comment.
  });
};
```

It produces this:
```js
import React from 'react';

const render = () => {
  return <div className="foo" />;
};
```

cpojer:
Please note we don't really take support requests for codemods but we happily will accept pull requests that improve them.

cc @TheSavior

alangpierce:
Ok. If it's ok, I'll send a PR to add a warning about the issue to the README. I also may eventually try modifying `create-element-to-jsx` to preserve comments, and I'll send a PR for that if I can get it working.

It might be good if you added a little more context to the README about the purpose and origin of the repo, how suitable the transforms are for general use, and how you approach external contributions/requests. My original impression was that the repo was somewhat "official", and that I could generally trust the transforms to be "safe" in the sense that behavior of the code doesn't change and all comments are preserved.

cpojer:
Sorry, we just don't have the volume to support this repo officially. It is mostly just a collection of react related codemods that we have run either internally at Facebook or that contributors have built. After we are done running a codemod, we usually don't need to run it a second time and we rely on the community to improve the codemods for their use cases.

If you'd like to see this fixed I'd recommend you to send a PR for it including some tests – I'm happy to review and merge it. Comments can usually be transferred simply by doing `newNode.comments = oldNode.comments;` :)

alangpierce:
No problem at all, I understand.

Thanks for the pointer! :-)

PR #60, opened 2016-06-28
Document bug in create-element-to-jsx, add Contributing section

alangpierce:
As mentioned in #59, the `create-element-to-jsx` codemod can destroy comments,
so leave a warning in the docs so people know to be careful.

Also add a "Support and Contributing" section at the bottom of the README that
briefly describes the state and purpose of the repo and how contributions are
handled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,10 @@ Converts calls to `React.createElement` into JSX elements.
 jscodeshift -t react-codemod/transforms/create-element-to-jsx.js <path>
 ```
 
+Note that this script will **destroy comments** within `createElement` usages,
+so you should manually check for any lost comments after running the
+transform.
+
 #### `findDOMNode`
 
 Updates `this.getDOMNode()` or `this.refs.foo.getDOMNode()` calls inside of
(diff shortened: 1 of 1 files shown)


cpojer:
I'd really prefer a fix in the codemod itself – are you working on this?

I'll merge this commit but am hoping for a follow-up PR.

alangpierce:
I'm taking a look, and hoping to get something working, although it's not really urgent/blocking for me at the moment.

reergymerej:
I may give it a go if you're not @alangpierce.  Let me know so I don't step on your toes.

alangpierce:
@reergymerej Sure, that would be great. I was mostly just investigating/learning things so far, and I probably wasn't going to continue on it until this weekend anyway.

A few things I noticed when I was initially taking a look at it, which may or may not be helpful:
* It looks like simply attaching `comments` to JSX nodes can cause the printer to produce code like `<span>/*comment*/</span>`, which isn't a valid comment. So it may be necessary to give some block comments their own `jsxExpressionContainer` node or something like that.
* I noticed that the JSX formatting was causing multiline `React.createElement` usages to be converted into single-line JSX. So it may be necessary to convert `//` comments to `/**/` comments (which I, personally, am fine with). But I'm not really sure how all that stuff works in jscodeshift/recast.

reergymerej:
Cool beans.  Can we open an "enhancement" issue so we don't lose track of this or do we need to do that in a fork?  Honestly, I'll forget all about it in a couple hours if there's no open issue.  :)

alangpierce:
@reergymerej FYI, I ended up implementing this, and sent out PR #61.

PR #61, opened 2016-07-04
Make create-element-to-jsx properly transfer comments

alangpierce:
Originally filed as #59.

Some non-obvious aspects of this change:

* For the main test, I put a comment between nearly every token. The generated
  output is really messy, but the main important thing is that every comment in
  the input appears somewhere in the output.
* For closing tags, I had to make a copy of the JSXIdentifier without any
  comments. Otherwise, `React.createElement('div' /*foo*/, null, ' ')` would
  become `<div /*foo*/> </div /*foo*/>`.
* I had to change the signature of `convertExpressionToJSXAttributes` to also
  return an `extraComments` value for any comments that didn't have an obvious
  attribute to attach to. This is also necessary since there might be no
  attributes, but still some comments.
* JSXText elements seem to ignore any comments you attach to them, so I had to
  just use the default behavior of a JSXExpressionContainer for any child
  strings that have comments.
* Attaching comments to JSXElement nodes that are children of other JSXElement
  nodes causes the comment to be incorrectly printed as a child string, e.g.
  `<div><span />/*a comment*/</div>`. To avoid this, whenever the script
  generates a child JSXElement, it wraps it in a JSXExpressionContainer if there
  are any comments.
* It looks like some comments can be neither leading nor trailing (e.g.
  `{/*foo*/}`, and that makes them not print when they're transferred to other
  nodes. So in some cases I had to explicitly set the `leading` and `trailing`
  booleans when transferring comments. This also makes the comment positioning
  look nicer in some situations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,10 +38,6 @@ Converts calls to `React.createElement` into JSX elements.
 jscodeshift -t react-codemod/transforms/create-element-to-jsx.js <path>
 ```
 
-Note that this script will **destroy comments** within `createElement` usages,
-so you should manually check for any lost comments after running the
-transform.
-
 #### `findDOMNode`
 
 Updates `this.getDOMNode()` or `this.refs.foo.getDOMNode()` calls inside of
--- a/transforms/__testfixtures__/create-element-to-jsx-preserve-comments.input.js
+++ b/transforms/__testfixtures__/create-element-to-jsx-preserve-comments.input.js
@@ -1,0 +1,38 @@
+var React = require('react');
+
+const render = () => {
+  return /*1*/React/*2*/./*3*/createElement/*4*/(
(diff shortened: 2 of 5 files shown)


alangpierce:
By the way, obviously it's out of scope for this change, but I bet jscodeshift/recast could be changed to automatically preserve comments, which would make this change simpler or unnecessary. For example, for any removed node with comments, it could find the lowest ancestor that still exists in the code and transfer the comment there. The printing step would also need to be better about never silently ignoring comments. Probably the behavior would need to be opt-in, but it seems like it would work well enough for most refactors, at least in the sense that it would guarantee that codemods never destroy documentation.

cpojer:
Thank you Alan!

Also cc @fkling in case he has some thoughts.

PR #65, opened 2016-07-08
inline single return statement without checking the type

keyanzhang:
With this change, it should transform 
```js
getDefaultProps: function() {
  return OtherClass.defaultProps;
},
```
to
```js
static defaultProps = OtherClass.defaultProps;
```

Reviewer: @spicyj


spicyj:
lgtm.

This also affects getInitialState right? I think that's okay.

keyanzhang:
@spicyj yeah, we scan for `this.something` usage separately so it should be fine.

PR #67, opened 2016-07-11
Initial version of manual bind to arrow function

vjeux:
This converts the manual bind to an arrow property initializer function.

It works for the base case `this.method = this.method.bind(this);`, two other patterns that I've seen in the codebase that should be codemodded as well in a subsequent version

- `(this: any).method = this.method.bind(this)`
- `const self = (this: any); self.method = this.method.bind(this)`

Test Plan:
npm t

I haven't tried it on the codebase yet, just want to get early feedback

### DIFF
--- a/transforms/manual-bind-to-arrow.js
+++ b/transforms/manual-bind-to-arrow.js
@@ -1,0 +1,182 @@
+/**
+ * Copyright 2013-2015, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+/**
+ * class Component extends React.Component {
+ *   constructor() { this.onClick = this.onClick.bind(this); }
+ *   onClick() { }
+ * }
+ *
+ * -->
+ *
+ * class Component extends React.Component {
+ *   onClick = () => { }
+ * }
+ */
+
+export default function transformer(file, api) {
+  const j = api.jscodeshift;
+
+  var root = j(file.source);
+
+  // Helper functions to transform a method declaration to an arrow function
+  // By default recast drops comments and jscodeshift doesn't have a way to
+  // set the return type in the convenience method. Otherwise we would have
+  // inlined all those.
+  function withComments(to, from) {
+    to.comments = from.comments;
+    return to;
+  }
+
+  function createArrowFunctionExpression(fn) {
+    var arrowFunc = j.arrowFunctionExpression(
+      fn.params,
+      fn.body,
+      false
+    );
+
+    arrowFunc.returnType = fn.returnType;
+    arrowFunc.defaults = fn.defaults;
+    arrowFunc.rest = fn.rest;
+    arrowFunc.async = fn.async;
+
+    return arrowFunc;
+  }
+
+  function createArrowProperty(prop) {
+    return withComments(j.classProperty(
+      j.identifier(prop.key.name),
+      createArrowFunctionExpression(prop.value),
+      null,
+      false
+    ), prop);
+  }
+
+  var hasChanged = false;
+  var transform = root
+    .find(j.AssignmentExpression)
+    .forEach(path => {
+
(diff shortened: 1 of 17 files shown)


transforms/manual-bind-to-arrow.js
```diff
@@ -0,0 +1,85 @@
...
+        path.node.right.callee.property.type === 'Identifier' &&
+        path.node.right.callee.property.name === 'bind' &&
+        path.node.right.callee.object.type === 'MemberExpression' &&
+        path.node.right.callee.object.property.type === 'Identifier' &&
+        path.node.right.callee.object.object.type === 'ThisExpression' &&
+        path.node.left.property.name === path.node.right.callee.object.property.name
+      )) {
+        return false;
+      }
+
```
keyanzhang: I think you can replace the above with
```js
  root
    .find(j.MethodDefinition, {
      kind: 'constructor',
    })
    .find(j.AssignmentExpression, {
      left: {
        type: 'MemberExpression',
        object: {
          type: 'ThisExpression',
        },
        property: {
          type: 'Identifier',
        },
      },
      right: {
        type: 'CallExpression',
        callee: {
          type: 'MemberExpression',
          property: {
            type: 'Identifier',
            name: 'bind',
          },
          object: {
            type: 'MemberExpression',
            object: {
              type: 'ThisExpression',
            },
            property: {
              type: 'Identifier',
            },
          },
        },
      },
    })
    .filter(path => 
      path.node.left.property.name === path.node.right.callee.object.property.name
    )
```

transforms/manual-bind-to-arrow.js
```diff
@@ -0,0 +1,85 @@
...
+              path.node.value.body
+            ),
+            null
+          )
+        );
+
+      return true;
+    })
+    .map(path => path.parentPath)
+    .replaceWith(path => null)
```
keyanzhang: If I recall correctly it might be better to do 
```js
.forEach(path => j(path).remove())
```
here. `.replaceWith(path => null)` leaves a semicolon here: https://astexplorer.net/#/kHE62wfbBC

transforms/manual-bind-to-arrow.js
```diff
@@ -0,0 +1,85 @@
...
+             (methodPath.node.type !== 'MethodDefinition' ||
+              methodPath.node.kind !== 'constructor')) {
+        methodPath = methodPath.parentPath;
+      }
+      if (!methodPath) {
+        return false;
+      }
+
+      if (!(
+        path.node.left.type === 'MemberExpression' &&
```
cpojer: this whole thing might benefit from some variable extraction like `left = path.node.left, right = path.node.right, callee = right.callee` etc. etc.

vjeux:
- Added support for:
  - `(this: any).method = this.method.bind(this);`
  - `const self: any = this; self.method = self.method.bind(this);`
- Now removes the constructor when there's no content left but a super call.
- It bails out early if there's nothing to transform.
- Added unit tests for all those.

I'm going to run this on www now and see how it looks. All the features I want are now supported, there are probably going to be things uncovered trying it out in the real codebase.

transforms/manual-bind-to-arrow.js
vjeux: Thanks for the trick! In the new version I've added conditionals inside of that logic so I'm not really sure we can do it as you mentioned.

Also, for the method lookup, I need the reference to that node, so I'd need to traverse it up anyway.

transforms/manual-bind-to-arrow.js
vjeux: Thanks! Done

transforms/manual-bind-to-arrow.js
vjeux: I like having it in one giant if, I can just comment stuff out and console.log things in the process. I'm not sure it would be really more readable or anything splitting them out in variables.

vjeux:
- Adding support for async functions
- Use the babylon parser to avoid issues with different line location from flow parser

vjeux:
Fixing lint error on one of the test that prevented travis to be green :)

keyanzhang:
Since this has already been battle tested let's merge this? 😃

PR #68, opened 2016-07-19
annotate unannotated non-liftable state with ExistsType

keyanzhang:
Reviewers: @spicyj 

Test Plan:
- It parses and works correctly (tested)

cc @samwgoldman


spicyj:
👍

PR #69, opened 2016-07-20
sort-comp instantiates regex from eslintrc properly

jozanza:
This fixes the issue where react/sort-comp eslint rules such as `"/^render+.$/"` are incorrectly converted to `/\/^render.+$\//` rather than `/^render.+$/`


jozanza:
fixes #70

cpojer:
This seems like it's breaking the tests.

jozanza:
@cpojer all better :)

PR #71, opened 2016-07-28
Bug fixes from running class transform at Facebook

spicyj:
`toString() {}` was getting converted to `static toString = () => {}`. JavaScript.

### DIFF
--- a/transforms/class.js
+++ b/transforms/class.js
@@ -285,7 +285,7 @@ module.exports = (file, api, options) => {
     }
     const invalidProperties = specPath.properties.filter(prop => (
       !prop.key.name || (
-        !STATIC_KEYS[prop.key.name] &&
+        !STATIC_KEYS.hasOwnProperty(prop.key.name) &&
         STATIC_KEY != prop.key.name &&
         !filterDefaultPropsField(prop) &&
         !filterGetInitialStateField(prop) &&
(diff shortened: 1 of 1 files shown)


spicyj:
@keyanzhang

transforms/class.js
```diff
@@ -285,7 +285,7 @@ module.exports = (file, api, options) => {
     }
     const invalidProperties = specPath.properties.filter(prop => (
       !prop.key.name || (
-        !STATIC_KEYS[prop.key.name] &&
+        !STATIC_KEYS.hasOwnProperty(prop.key.name) &&
```
benjamn: Can `prop.key.name` ever be `=== "hasOwnProperty"`?

spicyj: Yes, but STATIC_KEYS doesn't have that as a key.

keyanzhang:
lgtm, thanks for catching this!

cpojer:
Why not `Object.create(null)` for `STATIC_KEYS`? :)

`{}` is pure evil.

spicyj:
Why not a list? Why not a Set? Because this was easier from the code we had and I'm probably never going to look at this file again in my life.

spicyj:
And because while Object.create(null) would work for this use, sometimes it can trip you up with `__proto__` so I'm wary of using it and making people forget about their problems.

Issue #64, opened 2016-07-08
TypeError: Cannot read property 'start' of null in pure_component.js


ritz078:
Here's the error: 

```
TypeError: Cannot read property 'start' of null
    at reportSkipped (pure-component.js:86:28)
    at pure-component.js:97:9
    at Array.filter (native)
    at Collection.filter (/Users/ritz078/.npm-packages/lib/node_modules/jscodeshift/dist/Collection.js:60:46)
    at module.exports (pure-component.js:94:6)
```

keyanzhang:
Thanks for reporting this. Could you provide a sample of the code that caused this error?

ritz078:
```js
import React  from 'react';

export default class SingleAccordion extends React.Component {
  static propTypes = {
    actions  : React.PropTypes.object
  };

  constructor (props, context) {
    super(props, context);
    this.state = {
      expanded: false,
    };
    this.handleClick = this.handleClick.bind(this);
  }

  handleClick () {
    this.setState( { expanded : !this.state.expanded });
    if (this.props.trackClick && !this.state.expanded) {
      this.props.trackClick();
    }
  }

  render () {
    const Child = this.props.childElement ;
    const style = {} ;
    style.height = this.state.expanded ? 'auto' : this.props.height;
    const button = this.state.expanded ? this.props.expanded : this.props.collapsed ;

    return (
      <div>
        <div className='accordion-container f14black' style={style}>
          <Child content={this.props.content}/>
        </div>
        <div className="accordion-button" onClick={this.handleClick}>{button}</div>
      </div>
    );
  }
}
```

DrewML:
Huh. So this was/is failing because the `node` being returned for this `ClassDeclaration` is missing the `loc` object (it's `null`). Reproducible [here](http://astexplorer.net/#/PifJKAJ3we/1).

Could be fixed in this codemod itself by not reporting line/column information whenever it's missing (luckily this is only happening when the codemod tries to report that a `Class` was skipped), but this is probably an issue with the `babel-core` v5 parser that `jscodeshift` is using by default.

I tried stepping around the issue by specifying `--parser babylon` to `jscodeshift`, but `ast-types` then fails under the hood because `Babylon` diverges from the `ESTree` spec and uses the `StringLiteral` node type instead of the `Literal` type.

@keyanzhang Interested in a PR just to prevent throwing an exception, or you want to leave as-is?

keyanzhang:
[...]

PR #78, opened 2016-09-23
Prevented throwing exception in 'pure-component' transform

surgeboris:
see more at https://github.com/reactjs/react-codemod/issues/64


keyanzhang:
Thanks for the PR @surgeboris! Would you mind adding a descriptive `console.warn` before returning?

PR #79, opened 2016-09-24
Skip incorrectly-capitalized cases in create-element-to-jsx

alangpierce:
JSX uses capitalization to determine what type of `React.createElement` call to
produce: if the name starts with a lower-case letter, `React.createElement` is
called with the name as a string. Otherwise, it is called with the name as an
identifier. This means that an expression like `React.createElement(componentClass)`
cannot be converted to JSX, since `<componentClass />` would transform to the
string form. This commit changes the script to detect and skip these cases so
that the script does not introduce correctness issues.


cpojer:
thank you!

PR #83, opened 2016-09-28
Codemod to convert ES6 Class components to Stateless functional components

rosswarren:
Hi 😄 

I wrote a codemod that converts React ES6 class components to stateless functional components where possible.

It converts components that
- do not have any lifecycle methods other than render
- have either no constructor or a constructor that only calls `super()` or `super(props)`
- do not use refs

Also pulls statics out and add them as properties to the function component so that statics like propTypes are preserved.

Attempts to destructure props where possible.

Please let me know if you can think of any obvious holes in this. It worked very well for our project with a few hundred React components.

I am not really happy with the formatting of the destructuring parameters but I think that can't be helped unless some changes are made in recast. Also, I didn't account for flow. I need to learn it a bit first then I can add some tests for that.


sebmarkbage:
Does this cover passing `this` along anywhere such as `<Foo handler={this} />`?

Similarly, there is a use case you can't track from a file without finding all its callers.

`<Bar ref={component => doSomethingWithComponent(ReactDOM.findDOMNode(component))} />`

Even if a class has no methods on it, `findDOMNode` can be useful by getting a ref on it.

rosswarren:
@sebmarkbage thanks, I will look into these and add some more tests and get back to you

rosswarren:
@sebmarkbage I added a test cover the first case that you mentioned, however the second case as you say there is no way to track it 😞 I added a caveat to the Readme information. I think this codemod is still useful despite this caveat but will defer to your judgement!

eelyafi:
@rosswarren, thanks for the PR. I just used it. Noticed that it won't remove `Component` from `import React, { Component, PropTypes } from 'react';` would be great to remove it whenever it's not used anymore.

benjamn:
Removing unused imports could be a very useful standalone transform!

eelyafi:
@benjamn that's true. The reason why I mentioned this is because more likely no unused variables linter will fail if you don't remove unused Component, so you'll have to fix all the files manually.

rosswarren:
@eelyafi Good point, I think it would be better if this codemod did not leave you with unused imports. I didn't do this originally because I thought it would be cleaned up by eslint with `--fix` but it seems it is something that can't be. Will try to get it done today.

@benjamn I think that kind of general functionality would be really good to be contributed as part of eslint, since that is where you see the problem it would make sense if you could fix it there too.

eelyafi:
@rosswarren awesome, another thing that I noticed. When a component uses context, it should pass it as a second argument after props. Unfortunately codemod doesn't do this. Example:

```
// context won't be here
const Button = ({children}, context) =>
  <button style={{background: context.color}}>
    {children}
  </button>;
// this part is fine
Button.contextTypes = {color: React.PropTypes.string};
```

rosswarren:
@eelyafi Ah! I was not aware of this feature 👍 
I have updated with a test that covers the scenario you have described,

OliverJAsh:
Is this the same thing as the existing `pure-component` codemod? https://github.com/reactjs/react-codemod/tree/b69f98231088ad2f9a29608580f81873474c132e#pure-component

rosswarren:
@OliverJAsh looks like it!

zww930620:
where is utils/ReactUtils.... 
image @rosswarren

zww930620:
where is utils/ReactUtils.... 
image @rosswarren

PR #84, opened 2016-10-03
pure-component: support Type Annotations & Class Properties

danharper:
I was going to split this into a few PRs, but they're kinda related.

Firstly, it now treats any Component with a `ClassProperty` which isn't `props` as being impure.

Secondly, it retains any static `ClassProperty`s.

Finally, it retains `props` type annotations when transforming to a function.

I haven't used jscodeshift before, so happy to make any changes if there's a cleaner way of accomplishing it :)

``` js
// input
class PureWithTypes extends React.Component {
  props: { foo: string };
  render() {
    return <div className={this.props.foo} />;
  }
}

// output
const PureWithTypes = (props: { foo: string }) => {
  return <div className={props.foo} />;
}
```

### DIFF
--- a/transforms/__testfixtures__/pure-component.input.js
+++ b/transforms/__testfixtures__/pure-component.input.js
@@ -61,4 +61,20 @@ class ImpureClassPropertyWithTypes extends React.Component {
   }
 }
 
+class PureWithPropTypes extends React.Component {
+  static propTypes = { foo: React.PropTypes.string };
+  static foo = 'bar';
+  render() {
+    return <div />;
+  }
+}
+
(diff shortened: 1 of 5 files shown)


danharper:
Oh, this completely errors when running on my own project.. I guess the default parser of jscodeshift is Babylon, but not in the tests?

cpojer:
@keyanzhang can you take a look at this one? :)

keyanzhang:
Thanks @danharper, the code looks solid. I manually restarted the build on travis to double check; all the errors are from #88 but not `pure-component`, so this is good to merge!

danharper:
👍  thanks

PR #90, opened 2016-11-29
Class transform clean up

### LINKED ISSUE #88: class transform is broken
From the most recent [travis-ci build](https://travis-ci.org/reactjs/react-codemod/builds/173685402):
```
...
     [{kind: method, key: [object Object], value: [object Object], static: false, loc: null, type: MethodDefinition, comments: undefined, computed: false, decorators: null}] does not match field "body": [ClassMethod | ClassProperty] of type ClassBody
      
      at add (node_modules/jscodeshift/node_modules/recast/node_modules/ast-types/lib/types.js:580:31)
      at node_modules/jscodeshift/node_modules/recast/node_modules/ast-types/lib/types.js:593:21
      at Array.forEach (native)
      at Function.Object.defineProperty.value [as classBody] (node_modules/jscodeshift/node_modules/recast/node_modules/ast-types/lib/types.js:592:34)
      at createESClass (transforms/class.js:957:7)
      at NodePath.updateToClass (transforms/class.js:1068:5)
      at node_modules/jscodeshift/dist/Collection.js:76:36
      at Array.forEach (native)
      at Collection.forEach (node_modules/jscodeshift/dist/Collection.js:75:18)
      at apply (transforms/class.js:1137:264)
...
```

I saw the same failures when running `npm install` locally.

Looking at the [build history](https://travis-ci.org/reactjs/react-codemod/builds) it appears that https://github.com/reactjs/react-codemod/commit/84ef7b73b7abc21ee9ca537b59b53bbc0a690ac4 is good but https://github.com/reactjs/react-codemod/commit/509398000bf687b74ef560c8b3de38a6f70a9971 even though the latter didn't introduce any changes into class.js.

I thought maybe it was an issue with the version of jscodeshift I had so I installed 0.3.28 which is what the last passing build installed.  That did not fix the situation so I'm guessing it's some other dependency (or sub-dependency) that's causing the issue.

keyanzhang:
Fixes #88.

- All `class` tests are moved to a folder
- Fix specifiers definition (`id` -> `local`)
- Bump Jest
- Yarn it (and let it pick up benjamn/recast#340)
- Specify the new `objectCurlySpacing` print option (benjamn/recast#341)


cpojer:
Nice!

PR #91, opened 2016-12-04
Skip computed component values in create-element-to-jsx

alangpierce:
The create-element-to-jsx script was almost always calling `jsxIdentifierFor`,
which assumed that the expression consisted of strings, identifiers, and
non-computed member expressions (i.e. dot syntax rather than brackets syntax).
This ended up crashing or causing incorrect output when run on more dynamic code
like `React.createElement(a[b])`, `React.createElement(a[b.c])`, and
`React.createElement(f())`.


cpojer:
thanks!

PR #94, opened 2016-12-31
Pure component destructuring

andy-j-d:
Adds a ~~`destructure`~~ `destructuring` option to `pure-component` that will destructure the props where it's safe to do so. 
See the tests for use cases:
[Input](https://github.com/andy-j-d/react-codemod/blob/pure-component-destructuring/transforms/__testfixtures__/pure-component-destructure.input.js)
[Output](https://github.com/andy-j-d/react-codemod/blob/pure-component-destructuring/transforms/__testfixtures__/pure-component-destructure.output.js)


cpojer:
@keyanzhang is the man to review this.

keyanzhang:
Will do as soon as I find a stable internet connection!

--  
Keyan Zhang

transforms/pure-component.js
```diff
@@ -85,31 +86,83 @@ module.exports = function(file, api, options) {
...
-        j.variableDeclarator(
-          j.identifier(name),
-          j.arrowFunctionExpression(
-            [buildIdentifierWithTypeAnnotation('props', typeAnnotation)],
-            body
-          )
-        ),
-      ]
-    );
+  const build = ({ functionType }) => (name, body, typeAnnotation, destructure) => {
```
keyanzhang: nit: unnecessary to use a named parameter here

slorber:
hey.

Just a note as I already passed my codemod: I would have liked to use this option, but also why not using an arrow function instead of replacing the class by a real function?

andy-j-d:
@slorber if you use both the `destructure` (which will become `destructuring`) and `useArrows` options, you'll get destructured arrow functions.

slorber:
so bad I already did the migration, the useArrow option was not documented :)

andy-j-d:
I've put in those fixes @keyanzhang

andy-j-d:
@keyanzhang :ping:

keyanzhang:
Sorry for the late response -- somehow I missed the email notification. Thanks for the great work @andy-j-d!

PR #96, opened 2017-02-16
Don't put comments on JSXIdentifier nodes in create-element-to-jsx

alangpierce:
Previously, this code:

```js
React.createElement(
  Foo, // This is a comment
  {bar: baz}
);
```

would change to this code:

```js
<// This is a comment
Foo bar={baz} />
```

This is technically correct, but really ugly, so this commit changes the script
to reposition the comment to be on the outer JSXElement instead of the
JSXIdentifier directly. As we traverse the first arg to `React.createElement`,
we collect all comments and add them to the array of comments that we'll
eventually add to the outer JSXElement.

Note that this depends on a fix to recast that was just released in version
0.11.22. See this PR for more details: https://github.com/benjamn/recast/pull/364
Without that change, the repositioned comment could sometimes cause the
resulting code to be incorrect if it's at the start of the expression part of a
return statement. The latest jscodeshift has recast 0.11.22 in its version
range, so a fresh install makes everything work, but an old install might fail.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,14 +10,14 @@
     "prepublish": "npm run lint && npm run test"
   },
   "dependencies": {
-    "jscodeshift": "^0.3.25",
     "babel-eslint": "^6.0.5",
+    "babel-jest": "^15.0.0",
     "babel-plugin-transform-object-rest-spread": "^6.6.5",
     "babel-preset-es2015": "^6.6.0",
-    "babel-jest": "^15.0.0",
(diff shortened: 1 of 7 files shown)


alangpierce:
Build is failing because the build system doesn't have an updated recast. I actually just noticed that this repo uses yarn, so that should hopefully make that easy to fix. I'll see if I can fix by updating the yarn.lock.

cpojer:
Thanks!

PR #99, opened 2017-03-23
Migrate ReactNative View.propTypes to ViewPropTypes

bvaughn:
`ReactNative` will soon be deprecating `View.propTypes` in favor of `ViewPropTypes`. This codemod helps with that deprecation by doing the following:
* Updates all references to `View.propTypes` and replaces them with `ViewPropTypes`.
* Adds an `import` or `require` statement (depending on the convention followed in the parent file) for the new `ViewPropTypes`.
* Removes the `import`/`require` statement for `View` if it is no longer in use.
* Uses Haste modules (eg `import ViewPropTypes from 'ViewPropTypes';`) if it detects their use in the file else uses Node modules (eg `import { ViewPropTypes } from 'react-native';`).

### Testing

In addition to the newly added Jest tests this codemod was also run internally against Facebook's `ReactNative` code.

### Caveats

This codemod may introduce an unnecessary newline before certain types of imports. This is not a problem with the codemod but with recast. See facebook/jscodeshift/issues/185 and benjamn/recast/issues/371

This codemod doesn't handle the case demonstrated below:
```jsx
const ReactNative = require('react-native');
const { View } = ReactNative;
```

This pattern isn't common within Facebook and I didn't think it would be worth the extra complexity it would add to the codemod. If others feel strongly about this I could add support for it.

### DIFF
--- a/transforms/ReactNative-View-propTypes.js
+++ b/transforms/ReactNative-View-propTypes.js
@@ -20,7 +20,10 @@ const isReactNativeRequire = path => (
 
 const isRootViewReference = path => (
   path.node.name === 'View' &&
-  path.parent.node.type !== 'MemberExpression' &&
+  (
+    path.parent.node.type !== 'MemberExpression' ||
+    path.parent.node.object === path.node
+  ) &&
   (
     path.node.type !== 'JSXIdentifier' ||
     path.parent.node.type === 'JSXOpeningElement'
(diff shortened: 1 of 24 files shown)


acdlite:
🎉

keyanzhang:
We probably don't have this pattern internally so it's fine, but if `const View = require('View')` is the very first line of a program after some comments (like `@flow` or `@providesModule`) then replacing it will remove the comments ([example](https://astexplorer.net/#/gist/b8867eb1b42a9c53671af981bee296b2/9b49db9c58a2b77b0d6eb09b2644ed65019895c3)). Just FYI -- I have broken Haste a few times because of this :p

sebmarkbage:
In open source usage people aren't allowed to reference ViewPropTypes directly because CommonJS instead of Haste. Except in the internals of react-native. How is open source going to require this?

```js
require('react-native').ViewPropTypes
```

bvaughn:
@sebmarkbage Yeah, [I added a getter for `ViewPropTypes` to `react-native-implementation`](https://github.com/facebook/react-native/blob/master/Libraries/react-native/react-native-implementation.js#L126).

PR #101, opened 2017-03-28
Update class transform to fallback to react-create-class package

acdlite:
If React.createClass cannot be converted to a plain JavaScript class,
it falls back to using the react-create-class package instead.

### DIFF
--- a/transforms/__testfixtures__/class/class-initial-state.output.js
+++ b/transforms/__testfixtures__/class/class-initial-state.output.js
@@ -2,6 +2,8 @@
 
 import React from 'React';
 
+import ReactCreateClass from 'react-create-class';
+
 type SomeState = {foo: string};
 
 // only needs props
@@ -223,7 +225,7 @@ class DeferStateInitialization extends React.Component {
 
 var helper = () => {};
 
-var PassGetInitialState = React.createClass({ // bail out here
+var PassGetInitialState = ReactCreateClass({ // bail out here
   getInitialState() {
     return this.lol();
   },
@@ -237,7 +239,7 @@ var PassGetInitialState = React.createClass({ // bail out here
   },
 });
 
-var UseGetInitialState = React.createClass({ // bail out here
+var UseGetInitialState = ReactCreateClass({ // bail out here
   getInitialState() {
     return this.lol();
   },
@@ -251,7 +253,7 @@ var UseGetInitialState = React.createClass({ // bail out here
   },
 });
 
-var UseArguments = React.createClass({ // bail out here
+var UseArguments = ReactCreateClass({ // bail out here
   helper() {
     console.log(arguments);
   },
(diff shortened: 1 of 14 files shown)


transforms/__testfixtures__/class/class-pure-mixin3.output.js
```diff
@@ -0,0 +1,20 @@
...
+
+  getInitialState: function() {
+    return {
+      counter: this.props.initialNumber + 1,
+    };
+  },
+
+  render: function() {
+    return (
+      <div>{this.state.counter}</div>
```
keyanzhang: If the file uses JSX then we shouldn't prune `React` since we'll need `React.createElement` in scope

acdlite: Fixed

transforms/__testfixtures__/class/class-initial-state.output.js
```diff
@@ -223,7 +225,7 @@ class DeferStateInitialization extends React.Component {
 
 var helper = () => {};
 
-var PassGetInitialState = React.createClass({ // bail out here
+var PassGetInitialState = ReactCreateClass({ // bail out here
```
keyanzhang: Nit: should we call it `reactCreateClass` since it's a function?

acdlite: Wrote it this way because it's like a constructor, but don't care either way. Whatever convention we use should match https://github.com/babel/babel/pull/5554

keyanzhang:
Great work, thanks! LGTM, feel free to ping me if I can help when you run it internally

transforms/class.js
```diff
@@ -1103,23 +1118,80 @@ module.exports = (file, api, options) => {
...
+        findRequirePathAndBinding('React');
+
+      if (reactPathAndBinding) {
+        const {path, type} = reactPathAndBinding;
+        const noReactReferences = root
+          .find(j.Identifier)
+          .filter(path => path.value.name === 'React')
+          .length === 1;
+        const noJSX = root.find(j.JSXElement).length === 0;
+        const shouldRemoveReactImport = noReactReferences && noJSX;
```
keyanzhang: Oh, sorry I missed this part. If you run it on https://github.com/airbnb/enzyme/blob/master/src/ReactWrapperComponent.jsx you'll see the `PropTypes` import also gets removed.

This is only an issue when it's a `import` statement and has multiple specifiers on the left hand side. Let's add another guard for it?

acdlite: Good catch, thanks

transforms/__testfixtures__/class/class-initial-state.output.js
acdlite: Let's use `createReactClass` instead

PR #109, opened 2017-04-10
Fix issue with only named import of proptypes

relekang:
When creating reusable shapes it one only need proptypes and might
therefore not have imported react. This change make the proptype codemod
handle this use case.

Output of the added test before the change:
```js
import 'react';

export default PropTypes.shape({
  id: PropTypes.string,
})
```

Output of the added test after the change:
```js
  
export default PropTypes.shape({
  id: PropTypes.string,
})
```


bvaughn:
Thanks @relekang! There were a couple of merge conflicts but I resolved them and merged.

PR #115, opened 2017-04-10
Check if path.parent.node.init exists before reading its props.

mareksuscak:
Not sure why but it was failing to read the property in some cases such as:

```js
import React, { Component, PropTypes } from 'react'
```


tschaub:
See also #116 (I didn't see this when opening that).

mareksuscak:
@tschaub Yours is much more complete than mine incl. tests so go ahead and close mine.

bvaughn:
Sorry mate. Didn't see this. Already merged a similar fix (PR #127).

Going to close this as a dup but I really appreciate the effort!

PR #116, opened 2017-04-10
Ensure prop-types transform works with object pattern in arg

tschaub:
The `prop-types` transform assumes nodes of type `ObjectPattern` have a parent node with an `init` property.  This is not the case when an `ObjectPattern` is used in function params or call expression arguments (e.g. `function foo({bar}) {}` or `foo({bar})`).

It looks like this is meant to be working with a `VariableDeclarator` parent.  I've added a test that demonstrates the existing problem and updated the transform so the test passes.

### DIFF
--- a/transforms/React-PropTypes-to-prop-types.js
+++ b/transforms/React-PropTypes-to-prop-types.js
@@ -88,6 +88,7 @@ function removeDestructuredPropTypeStatements(j, root) {
   root
     .find(j.ObjectPattern)
     .filter(path => (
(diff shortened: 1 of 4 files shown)


tschaub:
It looks like #115 is trying to fix the same issue.  I'm happy to update this if you'd prefer to check `node.init` rather than `node.type`.

bvaughn:
Sorry mate. Didn't see this. Already merged a similar fix (PR #127).

Going to close this as a dup but I really appreciate the effort!

PR #117, opened 2017-04-10
Always add proptypes import for files with mixed imports & requires

robrichard:
If you have a file that uses both `import` and `require` you can end up with a broken file.

**Example**
```js
import React from 'react';
require('foo');

class MyClass extends React.Component {}

MyClass.propTypes = {
    foo: React.PropTypes.string
};
```

**Before** - Error: PropTypes is not defined.
```js
import React from 'react';
require('foo');

class MyClass extends React.Component {}

MyClass.propTypes = {
    foo: PropTypes.string
};
```

**After**
```js
import React from 'react';
require('foo');

class MyClass extends React.Component {}

MyClass.propTypes = {
    foo: PropTypes.string
};
```

This will add a PropTypes import if there is a React import or a PropTypes require if there is a React require

### DIFF
--- a/transforms/React-PropTypes-to-prop-types.js
+++ b/transforms/React-PropTypes-to-prop-types.js
@@ -39,13 +39,6 @@ const isReactPropTypes = path => (
   path.parent.node.object.name === 'React'
 );
 
-// Program uses ES import syntax
-function useImportSyntax(j, root) {
-  return root
-    .find(j.CallExpression, {callee: {name: 'require'}})
-    .length === 0;
-}
-
 // Program uses var keywords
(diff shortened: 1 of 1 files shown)


thepag:
Our styleguide uses sass, causing most of the repo to fall foul of this issue.
```
import React from 'react';
require('./my.scss');
```

bvaughn:
Hey @robrichard! Nice catch (and fix!)

Unfortunately it conflicts with (ad breaks) another recent fix for alpha-sorting. I've added the code you provided above as a new test though and fixed it with PR #128!

PR #121, opened 2017-04-12
PropTypes transform respects alphasort when inserting import/require

bvaughn:
Some people complained. 😅


bvaughn:
On a rainy day it would be useful to split a few of the things from this codemod into utils. I believe they'd be generically useful- specifically "insertImport" or "insertRequire" that respected alpha-sort.

PR #122, opened 2017-04-12
Remove trailing function comma in class transform

jsnmoon:
This trailing comma was causing the class transform to fail on my environment even when called with the `--babel` flag.

Environment info:
- macOS 10.12.2
- jscodeshift: v0.3.30
- babel and friends: v6.24.1

### DIFF
--- a/transforms/class.js
+++ b/transforms/class.js
@@ -1151,7 +1151,7 @@ module.exports = (file, api, options) => {
          ? j.callExpression(j.identifier(CREATE_CLASS_VARIABLE_NAME), [specPath])
          : j.callExpression(j.identifier(CREATE_CLASS_VARIABLE_NAME), classPath.value.arguments)
       ),
-      {comments},
+      {comments}
     );
   };
 


nickdima:
@jsnmoon I have the same problem. Do you think this will get merged or did you find a workaround? Other people don't seem to be having this problem.

jsnmoon:
Hi @nickdima, this issue occurs for me intermittently and I'm still in the middle of investigating why it occurs - no workarounds yet. My guess is that somewhere in our dependency chain, `babel-register` is loading a version of babylon that doesn't have `trailing-function-commas` plugin enabled by default.

No word on if/when this will be merged, unfortunately.

jsnmoon:
Hi @bvaughn, would it be possible to get this PR looked over? This, along with #132, #156, and #157 have been instrumental at leveraging this codemod over at @automattic.

Please let me know if you see anything amiss. I would be more than happy to address any outstanding issues on this PR as well as the other PRs mentioned above.

Thanks!

bvaughn:
Thanks for pinging me about these PRs! I'll take a quick look at the others.

PR #131, opened 2017-04-17
add import alias support for React-PropTypes-to-prop-types codemod

mkozmin:


bvaughn:
This looks great! Thanks!

I'll add onto it to support renames for requires too!

PR #133, opened 2017-04-20
Add codemod for 'React.DOM.div' -> 'React.createElement("div"'

flarnie:
Since we are deprecating the 'React.DOM.*' factories,(https://github.com/facebook/react/issues/9398 and https://github.com/facebook/react/pull/8356) we want to
provide a codemod so that it's easier for folks to upgrade their code.

This will include an option to use 'React.createFactory' instead of
'React.createElement' in case there is a use case where that is
preferred.

There is one use of `React.DOM.*` that I have seen which is not covered
here - sometimes it has mistakenly been used for Flow typing. In the
cases I have found it is not proper syntax and doesn't seem like
something we should cover with this codemod.

### DIFF
--- a/transforms/React-DOM-to-react-dom-factories.js
+++ b/transforms/React-DOM-to-react-dom-factories.js
@@ -1,0 +1,190 @@
+/**
+ * Copyright 2013-2015, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+'use strict';
+
+module.exports = function(file, api, options) {
+  const j = api.jscodeshift;
+  const root = j(file.source);
+
+  let hasModifications;
+
+  const DOMModuleName = 'DOM';
+
+  const isDOMSpecifier = specifier => (
+    specifier.imported &&
+    specifier.imported.name === DOMModuleName
+  );
+
+  /**
+   * Replaces 'DOM' with 'createElement' in places where we grab 'DOM' out of
+   * 'React' with destructuring.
+   */
+  const replaceDestructuredDOMStatement = (j, root) => {
+    let hasModifications = false;
+    //---------
+    // First update import statments. eg:
+    // import {
+    //   DOM,
+    //   foo,
+    // } from 'react';
+    root
+      .find(j.ImportDeclaration)
+      .filter(path => (
+        path.node.specifiers.filter(isDOMSpecifier).length > 0 &&
+        path.node.source.value === 'react'
+      ))
+      .forEach(path => {
+        hasModifications = true;
+
+        // Replace the DOM key with 'createElement'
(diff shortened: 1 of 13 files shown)


transforms/React-DOM-to-react-dom-factories.js
```diff
@@ -0,0 +1,180 @@
...
+
+  let hasModifications;
+
+  const DOMModuleName = 'DOM';
+
+  const isDOMSpecifier = specifier => (
+    specifier.imported &&
+    specifier.imported.name === DOMModuleName
+  );
+  const removeDeconstructedDOMStatement = (j, root) => {
```
flarnie: maybe a more accurate name would be 'replaceDeconstructedDOMStatement'.

In each case it replaces it with 'createElement'.

transforms/React-DOM-to-react-dom-factories.js
```diff
@@ -0,0 +1,180 @@
...
+          }
+        });
+      });
+    return hasModifications;
+  };
+
+  hasModifications =
+    removeDeconstructedDOMStatement(j, root) || hasModifications;
+
+  // is 'DOM'
```
flarnie: this comment probably was not needed

transforms/React-DOM-to-react-dom-factories.js
```diff
@@ -0,0 +1,180 @@
...
+        const DOMargs = path.parent.parent.parent.node.arguments;
+        const DOMFactoryPath = path.parent.parent.node.property;
+        const DOMFactoryType = DOMFactoryPath.name;
+
+        // React.DOM.div(... -> DOM.div(...
+        /*
+        j(DOMFactoryPath).replaceWith(
+          j.identifier('createElement') // TODO; fix
+        );
+        */
```
flarnie: Oops - removing this.

transforms/React-DOM-to-react-dom-factories.js
```diff
@@ -0,0 +1,190 @@
...
+      });
+    return hasModifications;
+  };
+
+  hasModifications =
+    replaceDestructuredDOMStatement(j, root) || hasModifications;
+
+  const isDOMIdentifier = path => (
+    path.node.name === DOMModuleName &&
+    path.parent.parent.node.type === 'CallExpression'
```
acdlite: This will match any function called `DOM`. Could we restrict this so it only matches if there's a corresponding import/require statement?

flarnie: Glad you are thinking of that!

The way I structured the codemod, it only gets used after we have already detected the corresponding import/require statement. I can restructure things to make that more clear.

That's also why I added [this test.](https://github.com/flarnie/react-codemod/blob/26f661fc51504f4291cd70027b4f7d7b2b4d2d03/transforms/__testfixtures__/React-DOM-to-react-dom-factories/react-dom-no-change-dom-from-other-libraries.input.js)

acdlite: Oh great, didn't see that!

PR #136, opened 2017-04-21
Add handling for nested destructuring

### LINKED ISSUE #134: `prop-types` codemod fails for explicit, nested destructuring
Given a `PropTypes` import that looks like this:

```javascript
const {
    PropTypes: {
        shape,
        bool,
        number,
        string
    }
} = React
```

the `prop-types` codemod will insert `import PropTypes from 'prop-types';` but will remove the destructuring assignment completely, leaving `shape`, `bool`, `number`, and `string` undefined.

mute:
Fixes #134 but not in a totally ideal way.

### DIFF
--- a/transforms/React-PropTypes-to-prop-types.js
+++ b/transforms/React-PropTypes-to-prop-types.js
@@ -105,7 +105,7 @@ module.exports = function(file, api, options) {
     throw new Error('No PropTypes import found!');
   }
 
-  // Remove PropTypes destructure statements (eg const { ProptTypes } = React)
+  // Remove PropTypes destructure statements (eg const { PropTypes } = React)
   function removeDestructuredPropTypeStatements(j, root) {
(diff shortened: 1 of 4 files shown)


transforms/React-PropTypes-to-prop-types.js
```diff
@@ -135,6 +139,15 @@ module.exports = function(file, api, options) {
           }
         );
 
+        // Add back any nested destructured children.
+        if (nestedPropTypesChildren) {
+          // TODO: This shouldn't just use a template string but the `ast-types` docs were too opaque for me to follow.
```
mute: This is the spot that could use some updating from someone with `jscodeshift` experience.

bvaughn: Those docs are a bit rough, huh.

PR #138, opened 2017-05-02
updates create class transform readme to use same path as other codemods

lukeschunk:
I noticed typically in the readme the instructions for the usage reference the path using `react-codemod/transforms/`. 

Was there a specific reason for this difference? If not, here as an update that makes it consistent.


threepointone:
Thanks for this, it's now changed to a nicer format.

PR #144, opened 2017-05-04
added AMD modules support to transform React-PropTypes-to-prop-types

rozenmanofer:


threepointone:
I don't think we'll support amd, but thank you (belatedly!) for the effort!

PR #145, opened 2017-05-14
Consistently use printOptions

jakezatecky:
The documentation makes reference to the ability to pass `--printOptions`, but this is not strictly observed in all transforms. Since these transformations can potentially affect hundreds of files, it might be annoying to not be able to change recast's defaults, like when it replaces all line-feeds with `CRLF` on Windows.

I tried to honor the existing format in each file with this PR. Since some conventions are not uniform among the existing transformations, I did not use the same format everywhere. Please let me know if you would like for me to adjust this PR as necessary.

I think it is worth noting that it may also be prudent to use something like `Object.assign` and merge the CLI and transform options, rather than just using the CLI options OR the transform's defaults.

### DIFF
--- a/transforms/React-DOM-to-react-dom-factories.js
+++ b/transforms/React-DOM-to-react-dom-factories.js
@@ -12,6 +12,7 @@
 
 module.exports = function(file, api, options) {
   const j = api.jscodeshift;
+  const printOptions = options.printOptions || { quote: 'single' };
   const root = j(file.source);
 
   let hasModifications;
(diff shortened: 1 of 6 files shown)


jakezatecky:
Also, it is not possible for the existing tests to pass on a Windows environment, because the output will always include `CRLF`, diverging from the fixtures with `LF`. I can further add `lineTerminator: '\n'` to each of these transformations, which will allow them to succeed on any OS.

threepointone:
Thank you for this!

PR #150, opened 2017-05-18
Fix React-PropTypes-to-prop-types for destructured require

trotzig:
I was using React-PropTypes-to-prop-types to convert files in our repo,
and I noticed that this type of construct wouldn't be converted:

  const { PropTypes } = require('react');

To fix this, I modified a function that handled constructs of the
following kind:

  const { PropTypes } = React;

Naming the test fixture was a little hard, since there are other
fixtures testing almost the same thing. I decided to use the term
"direct" to refer to a destructure done "directly" aimed at a require.


flarnie:
Very nice! Thanks for adding a test and for making this improvement. :)

PR #152, opened 2017-06-15
Retain leading comment fix #147

### LINKED ISSUE #147: Wrong PropTypes conversion if there are top comments
input:

```js
/**
 * Copyright 2013-2015, Facebook, Inc.
 * All rights reserved.
 *
 * This source code is licensed under the BSD-style license found in the
 * LICENSE file in the root directory of this source tree. An additional grant
 * of patent rights can be found in the PATENTS file in the same directory.
 *
 */
import React, { Component, PropTypes } from 'react';

class ClassComponent extends Component {
    static propTypes = {
        foo: PropTypes.string.isRequired,
    }
}
```

expected output:

```js
/**
 * Copyright 2013-2015, Facebook, Inc.
 * All rights reserved.
 *
 * This source code is licensed under the BSD-style license found in the
 * LICENSE file in the root directory of this source tree. An additional grant
 * of patent rights can be found in the PATENTS file in the same directory.
 *
 */
import React, { Component } from 'react';

class ClassComponent extends Component {
    static propTypes = {
        foo: PropTypes.string.isRequired,
    }
}
```

actual output:

```js
/**
 * Copyright 2013-2015, Facebook, Inc.
 * All rights reserved.
 *
 * This source code is licensed under the BSD-style license found in the
 * LICENSE file in the root directory of this source tree. An additional grant
 * of patent rights can be found in the PATENTS file in the same directory.
 *
 */
import React, { Component } from 'react';

class ClassComponent extends Component {
    static propTypes = {
        foo: PropTypes.string.isRequired,
    }
}
```

i took a look at the transform, seems this is an issue of jscodeshift or recast.

AugustinLF:
As shown in #147, if the React import is preceded by a comment, the `prop-types` import is printed before it, and not after. This is especially problematic when using the `// @flow` pragma, since it won't be a the top of the file anymore, and therefore won't work. This PR fixes this problem, using the [recipe advised by jscodeshift](https://github.com/facebook/jscodeshift/blob/master/recipes/retain-first-comment.md). However, I do have a whitespace problem, that I can't succeed to fix.

```javascript
/**
 * This is some multiline comment
 */
import React, { PropTypes } from 'react';

// gets converted to 
/**
 * This is some multiline comment
 */

import React from 'react';
```

If someone knows how to fix that, I'll update the PR ASAP. If not, I still do think that this PR could get merged as-is, since the line break should be less bothering than the misplaced comment.

### DIFF
--- a/transforms/React-PropTypes-to-prop-types.js
+++ b/transforms/React-PropTypes-to-prop-types.js
@@ -88,6 +88,16 @@ module.exports = function(file, api, options) {
           [j.importDefaultSpecifier(j.identifier(localPropTypesName))],
           j.literal(MODULE_NAME)
         );
+
+        // If there is a leading comment, retain it
+        // https://github.com/facebook/jscodeshift/blob/master/recipes/retain-first-comment.md
+        const firstNode = root.find(j.Program).get('body', 0).node;
+        const {comments} = firstNode;
+        if (comments) {
+          delete firstNode.comments;
+          importStatement.comments = comments;
+        }
+
         j(path).insertBefore(importStatement);
         return;
       }
--- a/transforms/__testfixtures__/React-PropTypes-to-prop-types/with-top-comment.input.js
+++ b/transforms/__testfixtures__/React-PropTypes-to-prop-types/with-top-comment.input.js
@@ -1,0 +1,11 @@
+/**
+ * This is some multiline comment
+ */
+import React, { PropTypes } from 'react';
+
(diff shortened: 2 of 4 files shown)


jsnmoon:
See https://github.com/benjamn/recast/issues/405 for why the extra line break is occurring.

bvaughn:
The whitespace problem can't be fixed for now, so far as I know. But this is a nice improvement. Thanks!

PR #154, opened 2017-06-26
Replace unordered list to ordered

Realetive:
Unordered list mean, that every item of list is optional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,14 +6,12 @@ APIs.
 
 ### Setup & Run
 
-  1. `yarn global add jscodeshift`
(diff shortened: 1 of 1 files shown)


bvaughn:
I don't think unordered lists imply "optional" but I do think they imply that the order isn't important. In this case, it is. However this is now conflicting. Would you mind updating the PR?

Realetive:
@bvaughn done. With numeric items we can say: «I have the problem on 3th step: *screenshot*». I think, it's more useful.

bvaughn:
Thanks~

PR #157, opened 2017-07-17
Class transform - Remove React import only if it should be removed

jsnmoon:
When replacing `React.createClass` with  `createReactClass`, the React import was being erroneously deleted even though JSX existed in the file.

This change ensures that the React import is not erroneously removed. Added a corresponding test case.


bvaughn:
Thanks for all of the recent fixes, @jsnmoon 😄

jsnmoon:
Glad to be of service! 🙂

PR #158, opened 2017-07-17
Added codemod for unstable_handleError -> componentDidCatch

bvaughn:
React introduced the experimental "error boundaries" feature in version 15. Components define an `unstable_handleError` method to opt into this feature. Version 16 finalizes support for error boundaries, but with a new name- `componentDidCatch`.

This codemod is pretty simplistic. It renames all identifiers named `unstable_handleError`. At first (a8be6bb) I thought about restricting renamed identifiers to class properties/methods and object expressions. But that felt fragile. (What if a react component directly called its own `unstable_handleError` method? We'd want to rename the call too.) So I broadened it (411750c). The broader approach runs the possibility of false positives but I think the name is sufficiently unique that it's okay.

Relates to the "make error boundaries official" beta-blocker in facebook/react/issues/8854

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,14 @@ Converts calls to `React.createElement` into JSX elements.
 jscodeshift -t react-codemod/transforms/create-element-to-jsx.js <path>
 ```
 
+#### `error-boundaries`
+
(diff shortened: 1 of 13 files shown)


bvaughn:
cc @flarnie for review 😄

PR #159, opened 2017-07-19
Fixed PropTypes codemod bug that left invalid variable references

bvaughn:
While running this codemod against some FB source I discovered a case that wasn't handled correctly:
```js
const React = require('react');
const ReactPropTypes = React.PropTypes;
const string = ReactPropTypes.string;
```
Would get incorrectly converted to:
```js
const React = require('react');
// ReactPropTypes is not defined
const string = ReactPropTypes.string;
```

This PR contains a fix as well as an added test case. Now the codemod only removes pointers that have names that conflict with the import, eg:
```js
const React = require('react');
const ReactPropTypes = React.PropTypes;
const string = ReactPropTypes.string;
```
Will now be converted to:
```js
const PropTypes = require('prop-types');
const React = require('react');
const ReactPropTypes = PropTypes;
const string = ReactPropTypes.string;
```

### DIFF
--- a/transforms/React-PropTypes-to-prop-types.js
+++ b/transforms/React-PropTypes-to-prop-types.js
@@ -208,8 +208,13 @@ module.exports = function(file, api, options) {
         hasModifications = true;
 
         // VariableDeclarator should be removed entirely
-        // eg 'PropTypes = React.PropTypes'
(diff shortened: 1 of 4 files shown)


flarnie:
I think this makes sense. Nice catch!

nice_cat_catch_attack

PR #166, opened 2017-08-01
Update `manual-bind-to-arrow` to support function-bind syntax

ryaninvents:
In the codemod, `this.foo = ::this.foo` will now be treated the same as `this.foo = this.foo.bind(this)`.

### DIFF
--- a/transforms/__testfixtures__/manual-bind-to-arrow/manual-bind-to-arrow9.input.js
+++ b/transforms/__testfixtures__/manual-bind-to-arrow/manual-bind-to-arrow9.input.js
@@ -1,0 +1,7 @@
+class Component extends React.Component {
+  constructor() {
+    super();
+    this.onClick = ::this.onClick;
+  }
+  onClick() { }
+}
--- a/transforms/__testfixtures__/manual-bind-to-arrow/manual-bind-to-arrow9.output.js
+++ b/transforms/__testfixtures__/manual-bind-to-arrow/manual-bind-to-arrow9.output.js
@@ -1,0 +1,3 @@
+class Component extends React.Component {
+  onClick = () => { };
+}
--- a/transforms/__tests__/manual-bind-to-arrow-test.js
+++ b/transforms/__tests__/manual-bind-to-arrow-test.js
@@ -21,6 +21,7 @@ var TESTS = [
   6,
   7,
   8,
+  9,
 ];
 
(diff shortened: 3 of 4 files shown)


evliu:
Wouldn't it normally be written as just `::this.foo` since that's the same as `this.foo = this.foo.bind(this)`?

r24y:
Yeah, but if you write `::this.foo` in your render method, it's the same as `this.foo.bind(this)`, which creates a new function every time it's rendered.

evliu:
In legacy code, we have it in our constructor

r24y:
Oh, I see. Simply writing `::this.foo` creates a new function but does not override the original.

Here's [an example](http://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=MYGwhgzhAECyCeBhA9gWwA7IHYFMsBdoBvAKGmmGwnwCcBXYfZGgCgEpizzp8ALASwgA6AGbJk0ALzQAXDL6DR4gNxdychcIBGYGqvIBfLlwhh4LVDihgA5jg6luFKshA4hIZDZYADABI4IJ7QAEQAJESW1nYGIT5s-tBG5FxiEuycTppCpuYA5Gl5CVzJ0Do00BmO3Nm5LHnlRYlGRiSUWNTOGFLQuADucEhomLgE7KrtnWk9lBhKyBNUhOUzw0LlqiRp4yTl40A&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=true&presets=stage-0&prettier=false&targets=&version=6.26.0&envVersion=) using the Babel REPL.

evliu:
ahh nice find

threepointone:
Did `::` ever make it beyond proposal stage?

ryaninvents:
> Did `::` ever make it beyond proposal stage?

I don't think so, but that's kind of why I wrote this 😄

We'd adopted `::` internally since we were overly confident that the syntax would get accepted. Later we couldn't migrate to `create-react-app` because of the experimental syntax. This codemod saved me a ton of time, since we had several affected repos.

I figure there must exist at least one other person or team who has painted themselves into the same corner as me 😆

threepointone:
Thanks much for the effort! I'm going to close this PR without merging because we don't want to support experimental syntaxes, but I'll remember it if the need does arise. Thanks again!

PR #172, opened 2017-09-13
Move everything to devDependencies

rgoldfinger:
I'm trying to add `react-codemod` as a dev dependency in our repo so that devs have easy access to current versions of these scrips without needing to install them seperately. However, I'm running into issues where the version of Jest in in this project is overriding our version of Jest, and our tests don't work. 

This fixes it by moving all `dependencies` to `devDependencies`. My understanding is that this repo should 'work' without anything installed since it a collection of scripts, and that all the dependencies are only for development. I'm happy to move some back to `dependencies` if I got that wrong.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,15 +10,6 @@
     "prepublish": "npm run lint && npm run test"
   },
   "dependencies": {
-    "babel-eslint": "^6.0.5",
-    "babel-jest": "^15.0.0",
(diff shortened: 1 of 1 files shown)


threepointone:
We moved a bunch of stuff into devDependencies for v5, but thank you for trying!

PR #174, opened 2017-09-22
Support the `type-annotations` group in sort-comp.

exKAZUu:
This PR supports the `type-annotations` group described in https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md


threepointone:
Thank you!

PR #176, opened 2017-10-06
Updated Readme to include npm to install

bytor99999:
I am not sure if the 3rd step also needs to have an npm option too, or if npm install means step 2 and 3 isn't needed. Thanks.


threepointone:
The instructions are a lot simpler now, but thank you!

PR #185, opened 2017-11-22
Fix jscodeshift arguments order

vfonic:
* `jscodeshift -t script.js <path>` doesn't allow `<path>` to be a directory, while, the correct invocation of the script from the [jscodeshift repo](https://github.com/facebook/jscodeshift) works:
`jscodeshift <path> -t script.js`


threepointone:
The instructions are a lot simpler now, but thank you!

PR #189, opened 2017-11-28
add a couple jsx tests and ensure they pass

knames:
added some ArrayExpression functionality for children, added some tests, and added a condition to check for empty second params
```
h('div', ['children'])
```


threepointone:
I assume this is a hyperscript to react transform? I don't think this should be in react-codemod, which is specifically for upgrades of react. Would recommend it be its own package instead. Thank you though!

PR #199, opened 2018-02-01
More broadly rename MemberExpressions

bvaughn:
Follow-up to #198

When running this codemod internally I found several additional places manually invoking lifecycles like `SomeComponent.componentDidMount.apply()` so the codemod has been updated to handle this case also.

So I've changed the codemod from renaming `CallExpression`s (eg `instance.method()`) to broader `MemberExpression`s (eg `instance.method()` _and_ `instance.method.apply()`).

Additional tests have been added.


gaearon:
Good catch. I’m so glad our names are greppable and not just onMount

bvaughn:
Oh my goodness, no kidding

PR #205, opened 2018-03-27
Added FAQ for commonly run into problem

ismusidhu:


threepointone:
added this in https://github.com/reactjs/react-codemod/pull/232, thank you!

PR #208, opened 2018-06-04
Added React 16.3 lifecycle methods to sort-comp

doytch:
The eslint `sort-comp` rule has been updated to include specific handling for the static React lifecycle methods introduced in v16.3. [Note the order in the lifecycle methods.](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md)

I ran this codemod today to bring our code up to spec and realised that it actually doesn't bring it up to spec with typical orderings and from the looks of it, won't [pass Airbnb's rule](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/rules/react.js#L226) which is the stated goal.

The issue is really the early-exit of the `static` check so I added an extra condition on that.

Additionally, I added the `UNSAFE_` methods for completeness. Not sure whether most people will be codemoding UNSAFE components but 🤷‍♂️


threepointone:
Could you rebase/merge from master? Also would be great with a test or two. Happy to land this once that's done.

doytch:
I can (and just did) rebase, but I'm on holiday for the next couple of weeks so can't add tests currently.

threepointone:
That's cool, I'll make a note to do it myself. Thank you very much for this!

PR #211, opened 2018-07-11
Updated so default order to include React ^16.3 lifecycle methods.

mrbinky3000:
Updated so that the default order includes the new React lifecycle methods.  Also, according to the latest eslint rules for React 16.3 and higher, static methods no longer always have to be at the top.  It depends on the name of the static method.  For instance, one of the new lifecycle methods, getDerivedStateFromProps, is a static method that must go under the constructor.  Static methods that start with get, set, must go under the lifecycle methods and before "everything else"


threepointone:
I merged https://github.com/reactjs/react-codemod/pull/208 instead, but thank you!

PR #215, opened 2018-09-15
Fix forgotten PureComponent in ReactUtils

bt4R9:
Hi guys.
I was implementing a `sort-comp` fixer in the project when noticed that PureComponent is always skipped. This happens because `findReactES6ClassDeclarationByParent` searches only for occurrences of a `Component` even if a `PureComponent` was passed.


threepointone:
could you add a test for this?

threepointone:
actually, it looks good to me, I'll merge it. Thank you!

PR #220, opened 2019-03-06
WIP: Add class to function transform

matthoffner:
Written around a few use cases like moving propTypes and defaultProps to the bottom of the file and including named props in the constructor.

Still needs some cleanup for readability and catches for not breaking on unsupported files.


threepointone:
We don't recommend converting class based components t oones with hooks unless super necessary. This issue has some recommendations, you could consider joining efforts there. Thank you very much for the effort!

matthoffner:
@threepointone Thanks for reviewing - I understand not switching over arbitrarily, my motivation for this was when eslint identifies stateless components that are classes https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prefer-stateless-function.md.

PR #223, opened 2019-04-16
POC - npx react-codemod <transform> <path/to/src>

threepointone:
An improvement to actually using these codemods. You should be able to `npx react-codemod <transform> path/to/src` and it should Just Work™️. Also updates the docs to prefer this version.

Did it as a quick hack so I could show during the meeting, happy to work on it further if it looks good. Would be good to point to in 16.9.0 release notes as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ npx react-codemod manual-bind-to-arrow <path>
 #### `pure-component`
 
 Converts ES6 classes that only have a render method, only have safe properties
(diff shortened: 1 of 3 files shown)


cpojer:
Oh my god it's so beautiful

kentcdodds:
YES PLEASE!

threepointone:
merging, so I can try it out. if it's broken, 🤷‍♂️

PR #229, opened 2019-07-18
upgrade jscodeshift, redo bin/npx experience, pass tests

threepointone:
To workaround an npx bug (which isn't even an issue in newer npx, sigh), I had to redo the cli experience. I copied it wholesale from jest-codemods, and rewrote it for react-codemods. Updating jscodeshift also broke some tests, which I fixed.


threepointone:
npx-codemod

threepointone:
I'll fix the build after some sleep 😴

PR #231, opened 2019-07-22
fix: add missing transformer option React-PropTypes-to-prop-types

### LINKED ISSUE #230: React-PropTypes-to-prop-types throws Invalid transform choice
Steps to Reproduce:
````
npx react-codemod React-PropTypes-to-prop-types src/js/components/*
````

Throws error:
````
Invalid transform choice, pick one of:
- rename-unsafe-lifecycles
- class
- create-element-to-jsx
- error-boundaries
- findDOMNode
- manual-bind-to-arrow
- pure-component
- pure-render-mixin
- react-to-react-dom
- ReactNative-View-propTypes
- sort-comp
````

I can see that https://github.com/reactjs/react-codemod/blob/master/bin/cli.js#L105` Does not contain `React-PropTypes-to-prop-types` option.

upadmdh:
`React-PropTypes-to-prop-types` was missing in the `TRANSFORMER_INQUIRER_CHOICES` list. This commit fixes that. 

Fixes #230

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -150,6 +150,11 @@ const TRANSFORMER_INQUIRER_CHOICES = [
   },
   {
     name:
+      'React-PropTypes-to-prop-types: Replaces React.PropTypes references with prop-types and adds the appropriate import or require statement',
+    value: 'React-PropTypes-to-prop-types'
+  },
+  {
+    name:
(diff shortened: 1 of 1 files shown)


threepointone:
Thanks for this! Will publish a patch version later today.

PR #228, opened 2019-07-14
Add TypeScript Support

skovy:
_Note: this is a branch off https://github.com/reactjs/react-codemod/pull/227 and it must be merged first._

## Motivation

Now that [`jscodeshift`](https://github.com/facebook/jscodeshift) supports TypeScript (added in [`v0.6.0`](https://github.com/facebook/jscodeshift/releases/tag/v0.6.0)) the `react-codemod` scripts can also support TypeScript without any custom logic.
 
Resolves https://github.com/reactjs/react-codemod/issues/206.

## Changes

- Upgrade to the latest version of `jscodeshift` to get parser support for TypeScript
- Update the `defineTest` helper in `jscodeshift`
    - The current helper doesn't allow a way to dynamically define the parser
    - The current helper assumes the file is always `.js` (not true for TypeScript)
- Fix tests to define the `parser` for those that need it (some flow tests that now break with the newer version of `jscodeshift`)
- Add basic tests for a TypeScript scenario for renaming deprecated lifecycle methods
- Update the `rename-unsafe-lifecycles` to also replace `ClassMethod` for TypeScript classes

## Testing

- [x] Tested against a TypeScript React public codebase: https://github.com/codesandbox/codesandbox-client/pull/2159
- [ ] https://github.com/reactjs/react-codemod/pull/227 and https://github.com/facebook/jscodeshift/pull/332 changes must first be merged, until then
    - Clone `jscodeshift` locally and checkout the branch with the necessary changes
    - In the `jscodeshift` directory `yarn link`
    - In this directory, `yarn link jscodeshift`
- [ ] `yarn test`

PR #236, opened 2019-07-27
Add TypeScript tests for the rename-unsafe-lifecycles transform

skovy:
Only the tests from https://github.com/reactjs/react-codemod/pull/228.

The transform has to be mocked because `jscodeshift` doesn't support testing other parsers/file extensions.

I named the examples `class.tsx.input.js` (must be `js` to match what `jscodeshift` expects) but still keep the `tsx` to help show that it's expected to actually be a TSX file (and can be if something like https://github.com/facebook/jscodeshift/pull/332 is merged). Happy to remove this and keep it only `class.input.js` 😕

### DIFF
--- a/transforms/__tests__/rename-unsafe-lifecycles-test.js
+++ b/transforms/__tests__/rename-unsafe-lifecycles-test.js
@@ -10,12 +10,6 @@
 
 'use strict';
 
-jest.mock('../rename-unsafe-lifecycles', () => {
-  return Object.assign(require.requireActual('../rename-unsafe-lifecycles'), {
-    parser: 'flow'
-  });
-});
-
 const tests = [
   'arrow-functions',
   'create-react-class',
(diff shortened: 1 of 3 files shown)


transforms/__tests__/rename-unsafe-lifecycles-test.js
```diff
@@ -30,12 +24,53 @@ const tests = [
...
+          require.requireActual('../rename-unsafe-lifecycles'),
+          {
+            parser: 'flow'
+          }
+        );
+      });
+    });
+
+    afterEach(() => {
+      jest.resetModules();
```
skovy: This was necessary to clear the `jest.mock` above, otherwise the following `typescript` block was ran with the `flow` parser because of the above mock. [Jest Docs](https://jestjs.io/docs/en/jest-object.html#jestresetmodules).

Thought about doing a `beforeAll`/`afterAll` since all the flow tests need the same mock, or promoting this `afterEach` to the top-level `describe`. Thought this was cleaner to run before/after every test and keep it specific to where it's needed but happy to change.

skovy:
@threepointone RE: https://github.com/reactjs/react-codemod/pull/228#issuecomment-515705420

Here's an attempt at adding test cases for TypeScript. It works but it's not elegant. Not sure if it's better to have no tests or these hacky tests (relying on `Object.assign` and file extensions that end with `.js`). Let me know what you think! 👍

threepointone:
Looks good to me! Thank you for doing this, appreciate the weekend effort.

skovy:
@threepointone definitely, thanks for the reviews! If any new transforms get added in future React releases feel free to tag me and I can test on a larger TypeScript codebase and add tests if needed 👍 For the most part the ASTs seem to be pretty similar but there was that one difference with the class methods. _(haven't tried any of the other transforms so those might have different issues)_

FDiskas:
```
npx react-codemod rename-unsafe-lifecycles --parser tsx
```
does not works in my case. All *.tsx files was skiped and `componentWillReceiveProps` was not renamed to `UNSAFE_componentWillReceiveProps`

For me looks like tests does nothing

threepointone:
@fdiskas please file a fresh issue with a reproducing example (like a git repo).

PR #237, opened 2019-07-30
Add support for handling groups for method ordering in sort-comp codemod

manixate:
Picking the logic from https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/sort-comp.js#L59 to support groups with order. Airbnb style guide also uses groups and react-codemod doesn't sort methods properly when defined using them.


threepointone:
I'll merge this and publish a new version later this week.

PR #247, opened 2019-11-01
Issues/181 : manual-bind-to-arrow does not clean up constructor binds when the same bind occurs multiple times

pksjce:
- Created a test with two components declared and bound functions with different names. This passes.
- Created a test with two components declared and bound functions with same name. This fails.

Added fix to scope the method/constructor removal to the corresponding component declaration instead of the root of the file.


threepointone:
Amazing, thank you very much! Nailed it.

pksjce:
Woah! that was quick! Thanks ❤️

PR #250, opened 2019-11-27
Check usage of `arguments` in methods

### LINKED ISSUE #163: Class.js codemod should fall back to create-react-class when a soon to become arrow function references "arguments" object.
Taken from [the MDN docs](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Arrow_functions#No_binding_of_arguments):

> Arrow functions don't bind an `arguments` object...

Thus _class.js_ transform should check whether or not a soon to become arrow function contains a reference to the `arguments` object. If it does, the codemod must fall back to using _create-react-class_ approach.

Not long after using the codemod, we ran into an issue with this function...

```js

closeAndCall(fn, event) {
  this.close()
  if (fn) {
    fn.apply(this, Array.prototype.slice.call(arguments, 1))
  }
  this.onClearSearchField()
}
```

... which has been transformed into the following piece of code:

```js
closeAndCall = (fn, event) => {
  this.close()
  if (fn) {
    fn.apply(this, Array.prototype.slice.call(arguments, 1))
  }
  this.onClearSearchField()
}
```

Now the _arguments_ var doesn't refer to the list of function arguments as before. Clearly, a solution, in this case, would be:

```js
closeAndCall = (fn, ...args) => {
  this.close()
  if (fn) {
    fn.apply(this, args)
  }
  this.onClearSearchField()
}
```

I might turn this into a PR eventually.

EDIT: 
For convenience, here's a babylon node visitor that finds the occurence:  
https://astexplorer.net/#/gist/2373f28729933bc7aef987a8cec35cd3/de71edf4ffb7e0d3ee390ebbfc3ab7a76739a939

jscodeshift version here: 
https://astexplorer.net/#/gist/2373f28729933bc7aef987a8cec35cd3/ffee55f7af7a3cf1820b907906b24713c3675b59

kgrz:
Fixes #163.

Although the issue mentions `class` transform, the problem likely was in
the `manual-bind-to-arrow` transform; `class` transform had checks to avoid
function transformation when `arguments` object was used in a method.

* Moved the `doesNotUseArguments` function into a separate util.


transforms/utils/doesNotUseArguments.js
```diff
@@ -0,0 +1,40 @@
...
+  const ReactUtils = require('./ReactUtils')(j);
+
+  const doesNotUseArguments = (path, filepath) => {
+    const hasArguments =
+      j(path)
+        .find(j.Identifier, { name: 'arguments' })
+        .size() > 0;
+
+    if (hasArguments) {
+      var warnStr = '';
```
kgrz: Not sure if there's a better way to do this :|

transforms/utils/doesNotUseArguments.js
```diff
@@ -0,0 +1,40 @@
...
+
+      if (filepath) {
+        warnStr += filepath + ': ';
+      }
+
+      warnStr += '`' + ReactUtils.directlyGetComponentName(path) + '` ' +
```
kgrz: The `directlyGetComponentName` outputs empty strings with/without these changes. Haven't dug in deep.

gaearon:
thanks!

PR #252, opened 2019-12-11
Fixes #51: Declare eslint dependency

### LINKED ISSUE #51: sort-comp is not working
Hi, when I run `sort-comp` transformation, it throws following error:
```
Error: Cannot read config package: eslint-config-airbnb
Error: Cannot find module 'eslint-config-airbnb'
Referenced from: C:\projects\client\.eslintrc
    at Function.Module._resolveFilename (module.js:440:15)
    at Function.Module._load (module.js:388:25)
    at Module.require (module.js:468:17)
    at require (internal/module.js:20:19)
    at loadPackage (C:\projects\react-codemod\node_modules\eslint\lib\config\config-file.js:168:16)
    at loadConfigFile (C:\projects\react-codemod\node_modules\eslint\lib\config\config-file.js:212:1
8)
    at load (C:\projects\react-codemod\node_modules\eslint\lib\config\config-file.js:385:18)
    at C:\projects\react-codemod\node_modules\eslint\lib\config\config-file.js:326:36
    at Array.reduceRight (native)
    at applyExtends (C:\projects\react-codemod\node_modules\eslint\lib\config\config-file.js:309:28)
```

I've run `npm install` within `react-codemod` directory but it didn't help.

laurens-dg:
This should fix the sort-comp code-mod as stated in #51

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
   },
   "dependencies": {
     "chalk": "^2.4.2",
+    "eslint": "^6.6.0",
     "execa": "^3.2.0",
     "globby": "^10.0.1",
     "inquirer": "^7.0.0",


laurens-dg:
I'd like to add @threepointone as a reviewer, but it seems like I would need write access to the repo according to the docs. So this tag and comment is my request for review.

threepointone:
Honestly, it's annoying that we need to include all of eslint for what's a very small piece of functionality. I think we can do better than this. https://github.com/reactjs/react-codemod/blob/6847d62ba3c89d3332e8574b16423bc423fdd418/transforms/sort-comp.js#L202-L206 I'll think about it some more, unless you have a better idea.

laurens-dg:
TBH the functionality is that we want to conform to eslint, it's only fair we check against the real thing.
Only improvement I can think of is marking it as a peer-dependency. 
This way we *should* be using the project's own eslint, but I don't know if this plays nice with npx.

threepointone:
Alright, I'm going to merge this, thank you for the PR! 

I'll do a release this weekend. We can revisit if there are complaints about download size.

vvscode:
@threepointone  just remind about the release ))

sayjeyhi:
@threepointone please consider to have a release

threepointone:
Done, released in v5.2.0. Give it a spin and lemme know how it goes?

PR #255, opened 2020-02-18
Fix pure component transform edge cases

wesleyyee:
**case 1: Default exported classes**
When converting a class combined with a default export, to a pure component, you end up creating a syntax error. This is because you cannot combine the default keyword with a lexical declaration. This PR separates the declaration from the export in this case. 

**case 2: Unused super class import remains**
When converting to a pure functional component the old super class may have a named import that is no longer used. This should be removed.

**case 3: Props argument exists even if its never used**

### DIFF
--- a/transforms/__testfixtures__/pure-component.input.js
+++ b/transforms/__testfixtures__/pure-component.input.js
@@ -31,6 +31,12 @@ class ImpureWithRef extends React.Component {
   }
 }
 
+class PureWithoutProps extends React.Component {
+  render() {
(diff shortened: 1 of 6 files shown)


wesleyyee:
@threepointone @bvaughn any thoughts?

threepointone:
At a glance it seems reasonable. Thank you for the PR! I'll do a proper review soon.

threepointone:
Thank you for your PR!

PR #259, opened 2020-04-09
Allow transformer parameters to be provided via cli

c-cal:


threepointone:
why is it important that the command accept these parameters via cli?

c-cal:
This PR addresses a personal need to eliminate human intervention. Was there another way to avoid the wizzard?

PR #266, opened 2020-07-06
Update update-react-imports with early return

lunaruan:
Sometimes, `jscodeshift` will modify files that it touched even though the codemod may not have actually modified any code. Because `jscodeshift` sometimes doesn't play well with types, this sometimes leads to files being modified unintentionally in strange ways. This PR adds an early return if we know we aren't going to modify any react imports in `update-react-imports` to avoid these unintentional modifications.

This PR also fixes a bug where we treat `import type React from 'react'` the same as `import React from 'react'`.

### DIFF
--- a/transforms/__testfixtures__/update-react-imports/react-type-default-export.input.js
+++ b/transforms/__testfixtures__/update-react-imports/react-type-default-export.input.js
@@ -1,0 +1,4 @@
+import type React from "react";
+import * as React from "react";
+
+<div>Hi</div>;
--- a/transforms/__testfixtures__/update-react-imports/react-type-default-export.output.js
+++ b/transforms/__testfixtures__/update-react-imports/react-type-default-export.output.js
@@ -1,0 +1,3 @@
+import type React from "react";
+
+<div>Hi</div>;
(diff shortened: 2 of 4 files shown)


threepointone:
I've merged this, but I'm having trouble publishing this. So I added you as a maintainer on https://www.npmjs.com/package/react-codemod. In your repo, run `npm version patch`, then `npm pub`, and then `git push`.

lunaruan:
Hey, I ran those commands and things seemed to work ok. LMK if I did it correctly or if there's something else I need to do!

threepointone:
yup, looks like the package got published correctly https://www.npmjs.com/package/react-codemod/v/5.2.2 but you also have to push the commit that's generated to this repo (ie - when you ran `npm version patch`, it makes a commit that looks like this https://github.com/reactjs/react-codemod/commit/c46e5f23763619201a82e9913766f1572bbbc49c, just push that to this repo directly)

threepointone:
`git push origin master`? or something like that

lunaruan:
Done! (I think)

threepointone:
yup, I see it https://github.com/reactjs/react-codemod/commit/cae173b9a425a68afa581c8770fe56f94efc97b1 yay you!

PR #268, opened 2020-07-17
Add name-default-component codemod

lfades:
**Note:** The codemod name is a wip

A codemod to add a camel cased name based on the file to anonymous functions and arrow functions. If the name already exists it appends `Component` to the name. Check the test cases to see the accepted cases and exceptions.

The main purpose for the change described above is to have an easier migration of components to support Fast Refresh.

Class components and `create-react-class` aren't the target of this codemod as it won't make a difference for Fast Refresh.

TODO: update readme


lfades:
Closing in favor of adding the codemod directly to Next.js, as we currently need it.

PR #271, opened 2020-09-09
update-react-imports: Destructure named exports

lunaruan:
This PR updates the `update-react-imports` codemod. Previously, this codemod removed unused React imports (as a result of the new JSX transform) and converted default imports to named imports. This PR goes a step further and destructures those named imports (ex. `import {useState} from 'react'` if they were previously default imports. Named React imports (`import * as React from 'react'`) are untouched

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ npx react-codemod sort-comp <path>
 
 #### `update-react-imports`
 
-[As of Babel 7.9.0](https://babeljs.io/blog/2020/03/16/7.9.0#a-new-jsx-transform-11154-https-githubcom-babel-babel-pull-11154), when using `runtime: automatic` in `@babel/preset-react` or `@babel/plugin-transform-react-jsx`, you will not need to explicitly import React for compiling jsx. This codemod removes the redundant import statements. It also converts  (`import React from 'react'`) to named imports (`import * as React from 'react`).
+[As of Babel 7.9.0](https://babeljs.io/blog/2020/03/16/7.9.0#a-new-jsx-transform-11154-https-githubcom-babel-babel-pull-11154), when using `runtime: automatic` in `@babel/preset-react` or `@babel/plugin-transform-react-jsx`, you will not need to explicitly import React for compiling jsx. This codemod removes the redundant import statements. It also converts  (`import React from 'react'`) to named imports (`import * as React from 'react'`).
 
 ```sh
 npx react-codemod update-react-imports <path>
--- a/transforms/__testfixtures__/update-react-imports/react-already-used-named-export.input.js
+++ b/transforms/__testfixtures__/update-react-imports/react-already-used-named-export.input.js
@@ -1,0 +1,3 @@
+import * as React from 'react';
+
+React.useState(false);
--- a/transforms/__testfixtures__/update-react-imports/react-basic-default-export-jsx-element-react-variable.output.js
+++ b/transforms/__testfixtures__/update-react-imports/react-basic-default-export-jsx-element-react-variable.output.js
@@ -1,5 +1,5 @@
-import * as React from "react";
+import { createElement } from "react";
 
-React.createElement('div', {});
+createElement('div', {});
 
 <div></div>;
--- a/transforms/__testfixtures__/update-react-imports/react-jsx-member-expression.output.js
+++ b/transforms/__testfixtures__/update-react-imports/react-jsx-member-expression.output.js
@@ -1,0 +1,3 @@
+import { Fragment, useState } from "react";
+
+<Fragment />
--- a/transforms/__testfixtures__/update-react-imports/react-not-removed.output.js
+++ b/transforms/__testfixtures__/update-react-imports/react-not-removed.output.js
@@ -1,0 +1,7 @@
+import * as React from "react";
+
+React.createElement('div', {});
+
+Promise.resolve(React);
+
+<div>Hi</div>
--- a/transforms/__testfixtures__/update-react-imports/variable-already-used.input.js
+++ b/transforms/__testfixtures__/update-react-imports/variable-already-used.input.js
@@ -1,0 +1,7 @@
+import * as React from "react";
+
+React.createElement('div', {});
+
+createElement('someFunction');
+
+<div>Hi</div>;
--- a/transforms/update-react-imports.js
+++ b/transforms/update-react-imports.js
@@ -11,7 +11,24 @@ module.exports = function(file, api, options) {
   const printOptions = options.printOptions || {};
   const root = j(file.source);
 
-  const reactImportPath = root
+  function isVariableDeclared(variable) {
+    return (
+      root
+        .find(j.Identifier, {
+          name: variable,
+        })
+        .filter(
+          path =>
+            path.parent.value.type !== 'MemberExpression' &&
+            path.parent.value.type !== 'QualifiedTypeIdentifier' &&
+            path.parent.value.type !== 'JSXMemberExpression',
+        )
+        .size() > 0
+    );
+  }
+
+  // Get all paths that import from React
+  const reactImportPaths = root
     .find(j.ImportDeclaration, {
       type: 'ImportDeclaration',
       source: {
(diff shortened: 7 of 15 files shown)


README.md
```diff
@@ -145,7 +145,7 @@ npx react-codemod sort-comp <path>
 
 #### `update-react-imports`
 
-[As of Babel 7.9.0](https://babeljs.io/blog/2020/03/16/7.9.0#a-new-jsx-transform-11154-https-githubcom-babel-babel-pull-11154), when using `runtime: automatic` in `@babel/preset-react` or `@babel/plugin-transform-react-jsx`, you will not need to explicitly import React for compiling jsx. This codemod removes the redundant import statements. It also converts  (`import React from 'react'`) to named imports (`import * as React from 'react`).
+[As of Babel 7.9.0](https://babeljs.io/blog/2020/03/16/7.9.0#a-new-jsx-transform-11154-https-githubcom-babel-babel-pull-11154), when using `runtime: automatic` in `@babel/preset-react` or `@babel/plugin-transform-react-jsx`, you will not need to explicitly import React for compiling jsx. This codemod removes the redundant import statements. It also converts  (`import React from 'react'`) to destructured named imports (`import { useState } as React from 'react`).
```
threepointone, line 148: This edit seems a bit oddly structured, might want to edit in (it's fine if it's in a future PR, not critical). Might not even need it, tbh.

transforms/__testfixtures__/update-react-imports/react-already-used-named-export.input.js
```diff
@@ -0,0 +1,3 @@
+import * as React from 'react';
```
threepointone, line 1: is the .output.js for this file meant to be empty? (I can't leave this comment on the output file because it's empty)

transforms/__testfixtures__/update-react-imports/react-jsx-member-expression.output.js
```diff
@@ -0,0 +1,3 @@
+import { Fragment, useState } from "react";
+
+<React.Fragment />
```
threepointone, line 3: Shouldn't this become `<Fragment/>`?

transforms/__testfixtures__/update-react-imports/react-basic-default-export-jsx-element-react-variable.output.js
```diff
@@ -1,5 +1,5 @@
-import * as React from "react";
+import { createElement } from "react";
 
-React.createElement('div', {});
+createElement('div', {});
```
threepointone, line 3: dope, love it.

transforms/__testfixtures__/update-react-imports/react-not-removed.output.js
```diff
@@ -0,0 +1,7 @@
+import * as React from "react";
+
+React.createElement('div', {});
```
threepointone, line 3: feels like the destructuring would still be useful here. Is the decision to bail completely?

transforms/__testfixtures__/update-react-imports/variable-already-used.output.js
```diff
@@ -0,0 +1 @@
+
```
threepointone, line 1: Was this meant to be empty output?

transforms/__testfixtures__/update-react-imports/variable-already-used.input.js
```diff
@@ -0,0 +1,7 @@
+import * as React from "react";
+
+React.createElement('div', {});
+
+createElement('someFunction');
```
threepointone, line 5: This hasn't been manually destructed so would likely fail in runtime, (or rather, the assumption will be that it's defined globally) Could you explain this testcase?

threepointone (commented):
I left some questions on the newly added tests, will review the actual code added to the transform once those are addressed, cheers 🍻

transforms/__testfixtures__/update-react-imports/react-already-used-named-export.input.js
lunaruan: Yeah. Empty output means the file won't be edited (we return null).

transforms/__testfixtures__/update-react-imports/react-not-removed.output.js
lunaruan: Yeah to simplify the code, we don't destructure if either React is already being used or if the variable we're destructuring to (ex. `createElement`) is already being used. Open to changing though.

transforms/__testfixtures__/update-react-imports/variable-already-used.output.js
lunaruan: Yup.

README.md
lunaruan: kay removed this edit for now

transforms/update-react-imports.js
```diff
@@ -11,7 +11,24 @@ module.exports = function(file, api, options) {
...
+          name: variable,
+        })
+        .filter(
+          path =>
+            path.parent.value.type !== 'MemberExpression' &&
+            path.parent.value.type !== 'QualifiedTypeIdentifier' &&
+            path.parent.value.type !== 'JSXMemberExpression',
+        )
+        .size() > 0;
+    return foo;
```
threepointone, line 27: trying to avoid 'foo' as a variable name 
```suggestion
    const variableReferences =
      root
        .find(j.Identifier, {
          name: variable,
        })
        .filter(
          path =>
            path.parent.value.type !== 'MemberExpression' &&
            path.parent.value.type !== 'QualifiedTypeIdentifier' &&
            path.parent.value.type !== 'JSXMemberExpression',
        );
    return variableReferences.size() > 0;
```

transforms/update-react-imports.js
```diff
@@ -23,28 +40,37 @@ module.exports = function(file, api, options) {
...
+  const reactPath = reactPaths.paths()[0];
+
+  const isDefaultImport = reactPath.value.specifiers.some(
+    specifier =>
+      specifier.type === 'ImportDefaultSpecifier' &&
+      specifier.local.name === 'React',
+  );
+
+  // Check to see if we should keep the React import
+  const reactImportInUse =
```
threepointone, line 73: been preferring 'predicate' style variable names for booleans lately. 
```suggestion
  const isReactImportUsed =
```

transforms/update-react-imports.js
```diff
@@ -54,16 +80,236 @@ module.exports = function(file, api, options) {
...
+        .size() === 0;
 
-  for (let i = 0; i < specifiers.length; i++) {
-    const specifier = specifiers[i];
-    if (specifier.local.name === 'React') {
-      if (specifier.type === 'ImportNamespaceSpecifier' && !keepReactImport) {
-        j(reactPath).remove();
+    if (canDestructureReactVariable) {
+      // If a variable that we are trying to import has already been declared,
+      // do not try to destructure imports
```
threepointone, line 116: What happens in the future when only the destructured imports work? Feels like we should at least drop a warning here that they need to fix it by hand or whatever. _Or_ we should re-alias the import and use a new named version instead.

transforms/update-react-imports.js
```diff
@@ -54,16 +80,236 @@ module.exports = function(file, api, options) {
...
+        .find(j.QualifiedTypeIdentifier, {
+          qualification: {
+            type: 'Identifier',
+            name: 'React',
+          },
+        })
+        .forEach(path => {
+          const id = path.value.id.name;
+          // This is a typeof import so it isn't actually a type
+          if (path.parent.parent.value.type === 'TypeofTypeAnnotation') {
```
threepointone, line 127: Move the comment inside the condition 
```suggestion
          if (path.parent.parent.value.type === 'TypeofTypeAnnotation') {
            // This is a typeof import so it isn't actually a type
```

transforms/update-react-imports.js
```diff
@@ -54,16 +80,236 @@ module.exports = function(file, api, options) {
...
+              canDestructureReactVariable = false;
+            }
+          }
+
+          if (variableAlreadyDeclared(id)) {
+            canDestructureReactVariable = false;
+          }
+        });
+
+      root
```
threepointone, line 146: please add a comment here to what this block of code does?

transforms/update-react-imports.js
```diff
@@ -54,16 +80,236 @@ module.exports = function(file, api, options) {
...
+
+          if (
+            variableAlreadyDeclared(property) ||
+            reactTypeIdentifiers[property]
+          ) {
+            canDestructureReactVariable = false;
+          }
+        });
+
+      root
```
threepointone, line 165: comment here too, please

transforms/update-react-imports.js
```diff
@@ -54,16 +80,236 @@ module.exports = function(file, api, options) {
...
+          ),
+        );
+      }
+      j(path).remove();
+    });
+  } else {
+    // Remove the import because it's not being used
+    // If we should keep the React import, just convert
+    // default imports to named imports
+    let importRemoved = false;
```
threepointone, line 302:
```suggestion
    let isImportRemoved = false;
```

threepointone (commented):
This is a fair bit of code, tried my best to review 😅 Please also make sure this works on typescript codebases, just as a sanity check.

transforms/__testfixtures__/update-react-imports/react-not-removed.output.js
lunaruan: I don't think the plan is to ever completely remove `React.createElement`. It'll still be there for people who want to manually create elements. We do plan to eventually stop letting the JSX transform use t.

transforms/update-react-imports.js
lunaruan: oh LOL oops I used `foo` so I could `console.log` the output and I forgot to change it >.>

transforms/update-react-imports.js
lunaruan: Named imports (both destructured and nondestructured) will always work. We are only planning on deprecating default imports.

PR #295, opened 2021-06-07
fix: spawn `jscodeshift` cli using node

merceyz:
**What's the problem this PR addresses?**

`react-codemod` tries to spawn sh files on Windows by spawning `node_modules/.bin/jscodeshift`

**How did you fix it?**

Resolve the JS CLI directly and spawn it using node

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -17,7 +17,7 @@ const chalk = require('chalk');
 const isGitClean = require('is-git-clean');
 
 const transformerDirectory = path.join(__dirname, '../', 'transforms');
-const jscodeshiftExecutable = require.resolve('.bin/jscodeshift');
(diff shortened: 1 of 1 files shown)


bin/cli.js
```diff
@@ -17,7 +17,7 @@ const chalk = require('chalk');
 const isGitClean = require('is-git-clean');
 
 const transformerDirectory = path.join(__dirname, '../', 'transforms');
-const jscodeshiftExecutable = require.resolve('.bin/jscodeshift');
+const jscodeshiftExecutable = require.resolve('jscodeshift/bin/jscodeshift.js');
```
Ayc0, line 20:
```suggestion
const jscodeshiftExecutable = require.resolve('jscodeshift/bin/jscodeshift');
```

merceyz: Correct, I was explicit to avoid some extra IO

PR #297, opened 2021-06-25
Don't rely on the .bin folder

arcanis:
The .bin folder isn't always guaranteed (it's an implementation detail that makes it easier for some package managers to add all relevant binaries to the PATH, but other approaches exist). Instead, packages should prefer accessing the binaries themselves from the path they list.

Fixes `yarn dlx react-codemod rename-unsafe-lifecycles`


arcanis:
Duplicate of https://github.com/reactjs/react-codemod/pull/295

PR #303, opened 2021-11-30
Update license in update-react-imports.js

eloytoro:
I assume this is accidental but given that this is an MIT-licensed repository all files under it should also be MIT no?


transforms/update-react-imports.js
```diff
@@ -1,7 +1,9 @@
 /**
- * (c) Facebook, Inc. and its affiliates. Confidential and proprietary.
+ * Copyright 2015-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
  *
- * @format
```
Ayc0, line 4: You should keep `@format`

PR #305, opened 2022-02-10
Use named components for type imports in update-react-imports

### LINKED ISSUE #283: update-react-imports adds extra import * from React
When an import is defined as

```
import React, { useState } from 'react'
```(usually due to automatically added imports)

The codemod transforms it to 
```
import { useState } from 'react';
import * as React from 'react';
```

Instead of just
```
import { useState } from 'react';
```

### PR DESCRIPTION
**Pull request by BPScott**
Fixes #283.

Previously running update-react-imports against

```tsx
import React from 'react';

const App: React.FunctionComponent<{ message: string }> = ({ message }) => ();
```

produced the following output:

```tsx
import * as React from 'react';

const App: React.FunctionComponent<{ message: string }> = ({ message }) => ();
```

Note that types are not converted to be named exports like when React.BLAH values are used, usage in types still gets  plucked off the namespace object. This PR applies the fix documented [here](https://github.com/reactjs/react-codemod/issues/283#issuecomment-814884712), so now it compiles to:

```ts
import { FunctionComponent } from 'react';

const App: FunctionComponent<{ message: string }> = ({ message }) => ();
```

### DIFF
--- a/transforms/__testfixtures__/update-react-imports/typescript/preserve-types-default.tsx.output.js
+++ b/transforms/__testfixtures__/update-react-imports/typescript/preserve-types-default.tsx.output.js
@@ -1,5 +1,5 @@
-import * as React from 'react';
+import { FunctionComponent } from 'react';
 
-const App: React.FunctionComponent<{ message: string }> = ({ message }) => (
+const App: FunctionComponent<{ message: string }> = ({ message }) => (
   <div>{message}</div>
 );
--- a/transforms/update-react-imports.js
+++ b/transforms/update-react-imports.js
@@ -31,6 +31,7 @@ module.exports = function(file, api, options) {
           path =>
             path.parent.value.type !== 'MemberExpression' &&
             path.parent.value.type !== 'QualifiedTypeIdentifier' &&
+            path.parent.value.type !== 'TSQualifiedName' &&
             path.parent.value.type !== 'JSXMemberExpression',
         )
         .size() > 0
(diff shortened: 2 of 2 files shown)


stropho:
I was looking for this right now. Well if it helps anyone, you can run the updated codemod from this PR before it gets merged (if ever) like this: 
```npx -p react-codemod@https://github.com/BPScott/react-codemod/tree/handle-type-imports#handle-type-imports react-codemod update-react-imports PATH```

stropho:
@lainermeister sorry, no idea. I tried to run the command right now and it works fine (Linux,Mac) even if I am in a directory where there is no `package.json` - (based on `npm can't find a package.json file in your current directory.`)
So my wild guess is that there was either a Github outage 🤣  or there are some security measures on your machine or network... Or the issues is completely somewhere else 🤷

TkDodo:
This PR correctly addressed 518 files out of 577 in our codebase that had React specific type imports, so thank you for that ❤️

see:

- https://github.com/reactjs/react-codemod/issues/325

I think it's fair to say that it "fixes" this issue, too.

alexbit-codemod:
thanks ben for this PR, which was used by @amirabbas-gh as the foundation for his fix: https://github.com/reactjs/react-codemod/issues/325#issuecomment-2547011607

i think we can close this out.

cc @rickhanlonii

BPScott:
Closing this. Fixed by https://github.com/codemod/commons/pull/9

PR #309, opened 2022-09-25
feat: Add codemod to transform string refs to arrow-functions

eps1lon:
We need to unfreeze `this.refs` first for this codemod to work. But can be used to sanity check first.

For https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#deprecate-string-refs-and-remove-production-mode-_owner-field

In: `<div ref="refComponent" />;`
Out: 
```jsx
<div
  ref={(current) => {
    this.refs['refComponent'] = current;
  }}
/>
```

Note that this codemod does not work without https://github.com/facebook/react/pull/25334

Should not be used if the codebase contains warnings about string refs without owners e.g.
```
Component "ComponentNameWithRef" contains the string ref "someStringRefName". 
Support for string refs will be removed in a future major release.
This case cannot be automatically converted to an arrow function.
We ask you to manually fix this case by using useRef() or createRef() instead.
Learn more about using refs safely here:
https://reactjs.org/link/strict-mode-string-ref
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,12 +175,6 @@ class ParentComponent extends React.Component {
     return (
       <div
         ref={(current) => {
-          if (process.env.NODE_ENV !== 'production') {
-            if (Object.isSealed(this.refs)) {
-              this.refs = {};
-            }
-          }
-
           this.refs["refComponent"] = current;
         }}
       />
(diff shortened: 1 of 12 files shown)


eps1lon:
We decided against a codemdoddable warning

eps1lon:
The plan is to land unfreezing `this.refs` for 18.3 and 19. Then we can use this codemod.

noahlemen (approved):

PR #310, opened 2022-12-02
Actualize sort-comp for supporting more selectors like static-variables, static-methods, getters, setters etc

DmitryKorlas:
This PR implements a following "selectors" from **react/sort-comp** [rule](https://github.com/airbnb/javascript/blob/eslint-config-airbnb-v19.0.4/packages/eslint-config-airbnb/rules/react.js#L243-L256):

 - static-variables
 - static-methods
 - instance-variables
 - getters
 - setters
 - instance-methods

Also it fixes an issues of sorting, i'll covered it by tests to make sure it works well.


stormwarning:
Any chance this could be merged? Having to manually patch this each time is kind of a pain 🙏

PR #319, opened 2024-03-01
feat: v19 codemods: removing memoization hooks, useContext to use, context initialization expression conversion, ref property access

r4zendev:
# React v19 codemods

## Summary

This PR contributes 4 new codemods to `react-codemods` repository, designed to streamline migration to a new React version that is going to have built-in compiler optimization features and a set of other semantic changes. Each codemod has its own description in the README file but below is a summary:

- React's memoization hooks removal. As per latest canary changelog, various optimizations will be handled by React compiler and there is not going to be a need to use memoization hooks anymore.
- React's `useContext` hook conversion to `use` hook.
- React's `<Context.Provider value={defaultContextValue}>` typical context initialization expression conversion to cleaner `<Context value={defaultContextValue}>`.
- React's `ref` property access passed via `forwardRef` function wrapper conversion to destructure the `ref` property from the `props` object, as per latest changelog.

### Codemod Platform

Codemod is an open-source platform that makes it easy for developers to build, share, and run codemods with a free-forever and open-source set of toolings such as Codemod Studio, CLI and VS Code Extension. Codemod platform offers an enhanced user experience, including simplified setup, execution, and access to extended list of features like the following:

- Ability to control include and exclude patterns by using glob-like pattern-matching
- Dry-running the codemods and integration with [VSCode extension](https://docs.codemod.com/deploying-codemods/vsce), which provides immediate access to diff in your editor's GUI
- Built-in formatting, error reporting, extensive developer experience, all in one place
- Access to dozens of other codemods built by community
- Access to these and other framework codemods via Codemod Registry, which is an npm-like registry for codemods. Codemods that are publish to the registry can be access via CLI, VS Code extension, and the web registry.
- The codemod platform is built to provide you with the most fine-grained control and the most insight about the impact of the codemods on your codebase. In the future, more analytics become available for codemod executions, performance and impact.

### Via Codemod CLI

- `npm install -g codemod` to install Codemod.com's official CLI application
- `codemod list` to locate the list of available codemods
- `codemod <codemod-name>` (for example: `codemod react/19/use-context-hook`) to run the codemod

Or just simply run:

```bash
npx codemod react/19/use-context-hook
```

By default, the current path where you run the codemod will be used as the target folder to be processed. For advanced configurations of Codemod CLI, please refer to the (https://docs.codemod.com/deploying-codemods/cli)

### Using react-codemods repository

```bash
npx react-codemod use-context-to-use <path>
```

Name should be specified as in the repository itself on GitHub.
For more details about how to run the codemods using the repository this PR is intended for, refer to the [README file in the root of the repository](https://github.com/reactjs/react-codemod?tab=readme-ov-file#usage).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-codemod",
-  "version": "5.4.3",
+  "version": "5.4.4",
   "description": "React codemod scripts",
   "license": "BSD-3-Clause",
   "repository": "reactjs/react-codemod",


eps1lon:
Are they for JS only or do they work on TypeScript files? If they do work on TypeScript files, can you include some tests for forwardRef specifically to showcase how type annotations are handled?

PR #339, opened 2025-08-19
fix(remove-forward-ref): preserve union and intersection types

iamakulov:
Previously, the codemod would drop union (|) and intersection (&) types from props when transforming `forwardRef` components, leaving components without proper type annotations:

```ts
// Before:
const Component = React.forwardRef<
  HTMLElement,
  { someProp: boolean } & { anotherProp: number }>(
  (props, ref) => ...
)

// Expected “after”:
const Component = (
  {
    ref,
    ...props
  }:
  { someProp: boolean }
    & { anotherProp: number }
    & { ref: React.RefObject<HTMLDivElement> }) => ...

// Actual “after”: types completely missing:
const Component = (
  {
    ref,
    ...props
  }) => ...
```

This PR fixes this, adding proper support for union and intersection types.

### DIFF
--- a/transforms/__testfixtures__/remove-forward-ref/typescript/intersection-types.input.js
+++ b/transforms/__testfixtures__/remove-forward-ref/typescript/intersection-types.input.js
@@ -1,0 +1,10 @@
+import React from "react";
+
+type MetaProps = { item: string };
+
+const Meta = React.forwardRef<
+  React.ElementRef<typeof Span>,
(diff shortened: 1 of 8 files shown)


iamakulov:
(Note: these tests will need to be updated a bit (`ref` → `ref?`) if https://github.com/reactjs/react-codemod/pull/338 lands.)

